PR #1, opened 2020-04-20
Sidebar

blangley28:
Implement the left sidebar on the builder screen. Created components for the buttons and the dropdown
Screen Shot 2020-04-20 at 1 54 35 PM


radamson:
I haven't had a chance to review yet, but the one thing I notice right off the bat is that we're missing the eslint/prettier configs from `pathways`.  We'll want to keep the styles consistent between these two repos.

For example, I see lots of changes from single to double quotes in the imports.

https://github.com/mcode/pathway-builder/blob/bcf68841a430bf58631c937d89f407bf2089b310/src/components/App.tsx#L2

dehall:
oh that's my bad. my copy must not have included all the dot files

src/components/DropDown/DropDown.module.scss
```diff
@@ -1,32 +1,28 @@
 @import '../../styles/variables';
 
 .dropdown {
```
noranda, line 3: In order to allow us to use other Material UI form elements without having to overwrite styles every time, I suggest moving the Material-UI specific files into overrides in the theme file.

To do so, delete all styles here except `width: 100%` and add the following to `src/styles/theme.tsx`:

```js
const materialUiOverrides = {
  // Style sheet name ⚛️
  MuiFormControl: {
    // Name of the rule
    root: {
      // Some CSS
      margin: '10px 0',
      '&:hover': {
        borderColor: colors.white
      }
    }
  },
  MuiFormLabel: {
    root: {
      color: colors.white,
      fontStyle: 'italic',
      '&$focused': {
        color: colors.white
      }
    }
  },
  MuiInputBase: {
    root: {
      color: colors.white,
      fontWeight: 600,
      '&:hover': {
        borderColor: colors.white,
        backgroundColor: colors.grayDark
      }
    }
  },
  MuiOutlinedInput: {
    root: {
      borderRadius: '0',
      '&:hover': {
        borderColor: colors.white
      },
      '&:hover $notchedOutline': {
        borderColor: colors.white
      },
      '&$focused $notchedOutline': {
        borderColor: colors.white
      }
    },
    notchedOutline: {
      borderColor: colors.white
    }
  },
  MuiSelect: {
    icon: {
      color: colors.white,
      fontSize: '2em'
    }
  }
};
```

Now add the following to the bottom of each theme in the same file:

```js
const theme = createMuiTheme({
  ...,
  overrides: { ...materialUiOverrides }
});

const projectorTheme = createMuiTheme({
  ...,
  overrides: { ...materialUiOverrides }
});
```

And finally, add this color at the top: `grayDark: '#444'`

src/components/DropDown/DropDown.tsx
```diff
@@ -1,59 +1,48 @@
-import React, { FC, useCallback } from 'react';
-import Select from 'react-select';
-
+import React, { FC, useState } from 'react';
 import styles from './DropDown.module.scss';
-import { Option } from 'option';
+
+import InputLabel from '@material-ui/core/InputLabel';
+import FormControl from '@material-ui/core/FormControl';
+import Select from '@material-ui/core/Select';
```
noranda, line 6: These 3 import lines can be condensed to:
```js
import { InputLabel, FormControl, Select, MenuItem } from '@material-ui/core';
```
(see `MenuItem` comment below)

src/components/DropDown/DropDown.tsx
```diff
@@ -1,59 +1,48 @@
...
-          onChange={onChangeCallback}
-          options={options}
-          aria-label={label}
-        />
-      </div>
-    );
-  else return <div></div>;
+  return (
+    <FormControl variant="outlined" className={styles.dropdown}>
+      <InputLabel htmlFor={id}>{label}</InputLabel>
```
noranda, line 30: Material UI [docs](https://material-ui.com/components/selects/) indicate we should use `id={id}` here instead of `htmlFor`.

src/components/DropDown/DropDown.tsx
```diff
@@ -1,59 +1,48 @@
...
+      <Select
+        native
+        value={selected.value}
+        onChange={setSelected}
+        label={label}
+        inputProps={{
+          id: id
+        }}
+      >
```
noranda, line 39: Per Material UI docs, change this to:

```js
<Select id={id} value={selected.value} onChange={setSelected} label={label}>
```

src/components/DropDown/DropDown.tsx
```diff
@@ -1,59 +1,48 @@
...
+        {options.map(option => (
+          <option value={option.value}>{option.label}</option>
```
noranda, line 41: We should be using Material UI's `MenuItem` component here. Change to:

```js
<MenuItem value={option.value}>{option.label}</MenuItem>
```

src/components/Sidebar/Sidebar.module.scss
```diff
@@ -0,0 +1,90 @@
...
+        border-collapse: separate;
+        border-spacing: 0 15px;
+
+        tr {
+          .button {
+            width: 40%;
+
+            button {
+              width: 90%;
+              margin-left: 5%;
```
noranda, line 72: Delete this line to be more in line with the design.

src/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,124 @@
+import React, { FC, useState, useEffect, useRef, RefObject } from 'react';
+
+import Button from 'components/Button';
```
noranda, line 3: Use Material UI's [`Button`](https://material-ui.com/components/buttons/#buttons-with-icons-and-label) component here.

noranda:
This is looking good! I have one more request. Can you add this to `materialUiOverrides` in the theme file?

```js
MuiButton: {
    root: {
      borderRadius: '0'
    },
    iconSizeMedium: {
      '&& svg': {
        fontSize: 14
      }
    }
  },
  MuiPaper: {
    rounded: {
      borderRadius: '0'
    }
  }
```

This will line up the styles a little better to the design. Otherwise this looks good to go!

src/styles/theme.tsx
```diff
@@ -30,6 +30,9 @@ const materialUiOverrides = {
     iconSizeMedium: {
       '& > *:first-child': {
         fontSize: 12
+      },
```
noranda, line 33: Lines 34-36 describe the same selector as lines 31-33. I think I like using `fontSize: 14` better, so lets change this section to:

```js
const materialUiOverrides = {
  MuiButton: {
    root: {
      borderRadius: 0
    },
    label: {
      fontWeight: 600
    },
    iconSizeMedium: {
      '& > *:first-child': {
        fontSize: 14
      }
    },
    iconSizeSmall: {
      '& > *:first-child': {
        fontSize: 14
      }
    }
  },
  ...
}
```

src/components/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,142 @@
...
+      <div className={styles.sidebarContainer} ref={sidebarContainerElement}>
+        <div className={styles.sidebar}>
+          <SidebarHeader currentNode={currentNode} />
+          <hr />
+          <DropDown
+            label={'Node Type'}
+            id={'Node Type'}
+            options={[
+              { label: 'Action', value: 'action' },
+              { label: 'Branch', value: 'branch' }
```
noranda, line 120: These should probably be consistent with our buttons. My design has "step" and "decision" for node types, terms I grabbed from another pathway builder. I'm ok with whatever terms we decide on, but they should be consistent. Currently our buttons show "action" and "decision".

dehall: I'd prefer to avoid the term "decision" because in a vacuum the software developers and clinicians expect opposite things from that term. Jim has referred to our action nodes as "decision nodes" because the clinician has to make a decision as to whether to accept or decline. "Branch" is less ambiguous.

src/styles/theme.tsx
blangley28: Updated.

src/components/Sidebar/Sidebar.tsx
blangley28: That's my bad I meant to bring that up to you. I went with 'Action' and 'Branch' because I was personally confused about 'step' and 'decision' and in design conversations we have always referred to them as action (or sometimes guidance) and branch nodes

noranda: I'm fine with "Action" and "Branch" so let's fix the buttons to match. Thanks!

blangley28: Updated the Decision button to say Branch

PR #2, opened 2020-04-23
Create new Pathway List View

mdube7:
Created the new Pathway List view.

The table is scrollable if you make the window small enough but I hid the scrollbar to keep it looking like the design.


src/components/PathwaysList/PathwaysList.tsx
```diff
@@ -1,174 +1,109 @@
...
+              </TableCell>
+              <TableCell>draft</TableCell>
+              <TableCell>2 days ago</TableCell>
+              <TableCell align="right">
+                <Button
+                  className={styles.editButton}
+                  color="primary"
+                  size="small"
+                  startIcon={<FontAwesomeIcon icon={faEdit} />}
+                >
```
dehall, line 57: Can we have this Edit button bring up the pathway view, the same way that the Select Pathway button previously did? I think it's only a couple line change -- add `onClick={() => editButton(pathway)}` here and down below where the PathwaysTable is rendered, change the editButton to `editButton={(pathway: Pathway): void => callback(pathway)}`

mdube7: Edit button will now show the pathway view

PR #4, opened 2020-04-30
Make Edit button on list load selected pathway

dehall:
This change makes the Edit button on the pathway list load the selected pathway, and shows the actual pathway title in the navigation bar.

While I was in there I removed a lot of the types and concepts that were only relevant to individual patient data, like the EvaluatedPathway, Documentation, etc. 

image


src/components/Graph/Graph.tsx
```diff
@@ -128,15 +105,15 @@ const Graph: FC<GraphProps> = memo(
...
-    }, [expandCurrentNode, evaluatedPathway.pathwayResults, setExpanded]);
+    // useEffect(() => {
+    //   if (pathway.pathwayResults) {
+    //     for (const currentNode of pathway.pathwayResults.currentStates) {
+    //       if (expandCurrentNode) {
+    //         if (currentNode) setExpanded(currentNode, true);
+    //       }
+    //     }
+    //   }
+    // }, [expandCurrentNode, pathway.pathwayResults, setExpanded]);
```
blangley28, line 116: Is there a reason this was commented out and not deleted? Is the intention to eventually use this again once we define a current node of the builder?

src/components/Graph/Graph.tsx
```diff
@@ -236,29 +203,18 @@ const GraphMemo: FC<GraphMemoProps> = memo(
...
-                      ? getPath(evaluatedPathway.pathwayResults).includes(nodeName) ||
-                        evaluatedPathway.pathwayResults.currentStates.includes(nodeName)
-                      : false
-                  }
-                  isCurrentNode={
-                    evaluatedPathway.pathwayResults
-                      ? evaluatedPathway.pathwayResults.currentStates.includes(nodeName)
-                      : false
-                  }
+                  isOnPatientPath={false}
```
blangley28, line 216: We should probably remove this prop all together (or at least rename it if it being used for some styling which we want to keep coupled with this prop)

src/components/Graph/Graph.tsx
```diff
@@ -288,11 +244,7 @@ const GraphMemo: FC<GraphMemoProps> = memo(
                     key={edgeName}
                     edge={edge}
                     edgeName={edgeName}
-                    isOnPatientPath={
-                      evaluatedPathway.pathwayResults
-                        ? isEdgeOnPatientPath(evaluatedPathway.pathwayResults, edge)
-                        : false
-                    }
+                    isOnPatientPath={false}
```
blangley28, line 247: Same here

src/components/App.tsx
```diff
@@ -40,48 +34,41 @@ const App: FC = () => {
...
 
-  const updateEvaluatedPathways = useCallback(
-    (value: EvaluatedPathway) => {
-      const newList = [...evaluatedPathways]; // Create a shallow copy of list
-      for (let i = 0; i < evaluatedPathways.length; i++) {
-        if (evaluatedPathways[i].pathway.name === value.pathway.name) {
+  const updatePathways = useCallback(
+    (value: Pathway) => {
+      const newList = [...pathways]; // Create a shallow copy of list
+      for (let i = 0; i < pathways.length; i++) {
```
blangley28, line 46: I know the list here isn't long but we should consider making the pathways state a map instead of a list

src/components/Graph/Graph.tsx
dehall: Yeah the idea was that we might want this piece again. But that's what source control is for isn't it. I'll delete it.

src/components/Graph/Graph.tsx
dehall: Will remove it. Based on the current designs all nodes should be blue except for the current node being edited, which is red, so I'll implement that here.

src/components/App.tsx
dehall: Upon closer review we can even delete this whole update function -- originally it was used when the ELM evaluation returned but now we don't need that anymore. I can still refactor the pathways state to be a map if you really want, but there's no longer any access to the list by name anymore.

dehall:
Updated based on feedback - the only thing left is there are still a lot of styles for "onPatientPath" vs "notOnPatientPath" which I'd still like to clean up, or we can push that to another task.
Also I'm not sure why it thinks the entirety of Graph.tsx changed

dehall:
Updated styles as well. Ready for re-review

dehall:
Also if you view with whitespace changes disabled the Graph delta looks a lot cleaner. Not sure what it thinks the whitespace changes are

src/components/PathwaysList/PathwaysList.tsx
```diff
@@ -94,9 +94,9 @@ const PathwaysList: FC<PathwaysListProps> = ({ evaluatedPathways, callback, serv
       </Button>
 
       <PathwaysTable
-        pathways={evaluatedPathways}
+        pathways={pathways}
         deleteButton={(): void => callback()}
```
blangley28, line 98: Since the delete button is not implemented yet can we remove the `callback()` call. This call triggers the screen to change but it isn't passing in a pathway. I think for now we should just have this do nothing
```suggestion
        deleteButton={(): void => {}}
```

blangley28: Sorry I forgot lint doesn't want it to be empty
```suggestion
        deleteButton={(): void => {
             // do nothing
        }}
```

PR #5, opened 2020-05-05
Sidebar link

blangley28:
When editing a pathway clicking on a node will select it. This sets the currentNode state in the App which is used by the Sidebar. Currently the only thing which changes in the sidebar is the name of the node.


src/components/Graph/Graph.tsx
```diff
@@ -92,13 +98,31 @@ const Graph: FC<GraphProps> = memo(({ pathway, interactive = true, expandCurrent
...
+        // selected a different but already opened node
+        return { ...prevState, expandedArray: prevState.expandedArray, lastSelectedNode: key };
+      } else {
+        // selected the same node or a different but collapsed node
+        const newExpandedArray = {
+          ...prevState.expandedArray,
+          [key]: !prevState.expandedArray[key]
+        };
+        return { ...prevState, expandedArray: newExpandedArray, lastSelectedNode: key };
+      }
```
dehall, line 125: It looks like the first and third blocks are exactly the same, and all 3 instances return the same structure, so could this be reduced to:
```suggestion
      let newExpandedArray = prevState.expandedArray;

      // if clicking on a node that was not already expanded,
      // or re-clicking on the current already open node, toggle it
      if (!prevState.expandedArray[key] || prevState.lastSelectedNode === key) {
        newExpandedArray = {
          ...prevState.expandedArray,
          [key]: !prevState.expandedArray[key]
        };
      }

      return { ...prevState, expandedArray: newExpandedArray, lastSelectedNode: key };
```

src/components/Graph/Graph.tsx
```diff
@@ -92,13 +98,31 @@ const Graph: FC<GraphProps> = memo(({ pathway, interactive = true, expandCurrent
     }, {});
   }, [layoutKeys]);
 
-  const [expanded, _setExpanded] = useState<{ [key: string]: boolean | undefined }>(
-    initialExpandedState
-  );
+  const [expanded, _setExpanded] = useState({
+    expandedArray: initialExpandedState,
```
dehall, line 102: Can we call this something else? It's really more of a map than an array

blangley28:
> Overall looks good. It would be really nice if the logic in `setExpanded` could leverage the `currentNode` from the PathwayContext rather than defining its own additional field for that, but maybe that's crossing too many React streams to do cleanly.

Thats the issue I was fighting. If someone could figure that out I will gladly update but this was the only solution which worked

dehall:
We had some discussion about this in the slack channel on Tuesday as well -- I'm good with either the current approach, or with the approach you're describing @noranda which I interpret as "only the current node can be open". Both have downsides when switching back & forth between nodes so I don't have a strong feeling either way.

Hover can be tricky to do well, but I could see a couple different ways it might work: 1 is that you have to hover for like a second to actually open the node and there's a little indicator on the cursor. I can think of some video game examples like this. (Hold X to open door) Option 2 is to expand a collapsed node immediately on hover, but not re-render the whole graph, and then immediately hide on mouse-out. Is that easy to implement though? If it's really complex I'd want to hold off until we had more solid user feedback

blangley28:
I agree if we use hover it should be at least a second of hover just so it doesn't go crazy. In my personal opinion we should leave the behavior as it is for right now and create a task to do some user research. I agree a different design is needed but instead of constantly changing it now I think we should come up with a solid plan and then have a small ticket in the future to implement it.

noranda:
@dehall I was thinking more along the lines of option 2. I don't foresee it being too difficult to implement with `onMouseOver` and `onMouseOut` event handlers, but I've also never implemented anything like this so I'm not sure. I don't think this should hold up this PR, but we could create a new task to either explore these options and/or implement them.

dehall:
I created PATHWAYS-257 for the interaction question

PR #7, opened 2020-05-09
Code cleanup and restructure

noranda:
This PR came about as I was reviewing previous PRs and noticed that the repo has become somewhat difficult to reason about for various reasons, but primarily because we haven't been using contexts correctly. This also causes issues with re-rendering that can become bigger issues down the road, so this PR is an attempt to do some basic restructuring to reset our codebase foundation in order to make future tasks easier to implement. Below is an overview of the changes I made.

### :white_check_mark: Simplified `App.tsx`
- A lot of the logic in `App.tsx` belonged in the Pathways context including using the service to populate pathways, so this was moved.
- The router was relocated here to create separate routes for our list of pathways and our builder. This makes it possible to use routing to store things like the selected pathway and current node, which in turn makes it possible for the user to refresh without losing their place.
- The `BuilderView` component was moved to a new `Builder` component which handles displaying the navigation, sidebar, and graph if a pathway is selected. Moving the navigation into this component removed it from the artifact list view, which we didn't need.
- Memoized to limit rerenders.

### :white_check_mark: Created `BuilderRoute.tsx`
- Created this route to handle routing the user to a given pathway and current node in that pathway. Before, if the user refreshed the page, they would lose their place in the workflow. This also means we no longer need to keep track of state for selected pathways and current nodes and it provides sharable urls for users.
- Because we don't yet have an API, the pathway id has been set to the json filename.
   Ex: ```http://localhost:3000/#/builder/neoadjuvant_pathway.json/```
   This can be changed to an id or another identifier in the future.
- The node id is set to the JSON key for the node state, which should be unique. Again we can change this once we have an API set up.
   Ex. ```http://localhost:3000/#/builder/neoadjuvant_pathway.json/node/TumorSize```

### :white_check_mark: Updated `PathwaysProvider.tsx`
- While this context provider existed, we were not using it correctly and in many cases we were unnecessarily prop drilling, which is what using contexts solves. For example, in `App.tsx`, we were pulling in the pathways and then passing them down to child components instead of storing them in the context. 
- Removed `pathway` and `currentNode` fields which are now stored in the route url.
- Added `pathways` and associated logic to use the service to pull in all pathways.

### :white_check_mark: Updated `PathwayList.tsx`
- This component was pretty large and complex, so it was broken down into separate components, but still living in the same folder since they are closely related (`PathwayList.tsx`, `PathwayTable.tsx`, and `NewPathwayModal.tsx`)
- Material UI styles were removed from the component and, where appropriate, combined with the stylesheet as `styles.tsx` to prevent having styles live in different locations. We did create a task to fix this everywhere in the app, so I only fixed these in components I looked at. Refer to these components when that task is assigned.
- In `NewPathwayModal.tsx`, removed state and used refs instead to prevent rerenders every time the inputs are changed. Also added an id to the pathway and reformatted to use the correct Material-UI elements.
- Moved `addPathway()` to Pathways provider and simplified.
- Added loading component where appropriate.
- Memoized all components and used `useCallback` hooks to limit rerenders.

### :white_check_mark: Updated `Navigation.tsx`
- Styles were combined as in `PathwayList.tsx`.
- Added a navigation link back to the table view.
- The Navigation component will no longer render in the pathways list view.

### :white_check_mark: Updated `Header.tsx`
- Moved the logo import into this component so it doesn't need to be passed in.
- Added a link on the logo to redirect back to the pathways list.
- Memoized all components and used `useCallback` hooks to limit rerenders.

### :white_check_mark: Updated `Graph.tsx`
- Updated to use the url node ids for keeping track of current nodes.
- Simplified expanding states logic. A future task will look into improving how and when nodes should expand.
- This component is still large and complex and should be restructured in the future.

### :white_check_mark: Updated `Sidebar.tsx`
- Minor code cleanup.
- Memoized all components to limit rerenders.
- We will likely need to break this down into smaller components in a future task.

### :white_check_mark: Updated `ThemeProvider.tsx`
- Updated `theme.tsx` and the theme provider to include a light and dark theme. This made it easier to style the new pathways modal, which uses a light background, and keep the styles for the sidebar, which uses a dark background.
- Updated the toggle theme function to work with both light and dark themes, although we may want to create a separate dark projector theme in the future.

I realize this is a large PR with some significant changes. If you see anything here that looks off, let me know. I suggest we merge this in before any other PRs, and I can help with any rebasing on current branches if needed.


src/components/Graph/Graph.tsx
```diff
@@ -90,30 +89,24 @@ const Graph: FC<GraphProps> = memo(({ interactive = true, expandCurrentNode = tr
...
-          ...prevState.expandedMap,
-          [key]: !prevState.expandedMap[key]
-        };
-      }
-      return { ...prevState, expandedMap: newExpandedMap, lastSelectedNode: key };
-    });
+  const toggleExpanded = useCallback((key: string) => {
+    setExpanded(prevState => ({
+      ...prevState,
+      [key]: !prevState[key] || prevState.lastSelectedNode === key,
```
dehall, line 106: It's not immediately obvious to me why the behavior is different, but I think it's this line -- switching to an expanded node "A" from a node that was just collapsed on the previous click "B" now collapses node A instead of just switching to it.
Steps:
1. Click on a node A to open & select it
2. Click on a node B to open & select it, A is still open
3. Click on B again to close it, A is still open
4. Click on A again.
Before this would leave A open and select it, now it collapses A.
We do have a task in the backlog (257) for reviewing the interaction in more detail so it's not critical, but if we can change one line and bring it back to how it was before that would be great

noranda: Good catch! I just pushed up the fix.

src/components/PathwaysList/NewPathwayModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const styles = useStyles();
+  const pathwayNameRef = useRef<HTMLInputElement>(null);
+  const pathwayDescRef = useRef<HTMLInputElement>(null);
+  const { addPathway } = usePathwayContext();
+
+  const createNewPathway = useCallback(
+    (event: FormEvent<HTMLFormElement>): void => {
+      event.preventDefault();
+
+      addPathway({
```
blangley28, line 33: I noticed this no longer goes directly to the BuilderView. I feel like once a user creates a new pathway they would always want to begin building it. But now after creating it they have to select edit to open the builder view

src/components/Graph/Graph.tsx
blangley28: I'm still observing some weird behavior with the nodes. With two nodes open it seems to work fine but with more than 2 nodes open sometimes switching to a node will select it and close it. Haven't taken a look at the code to see why but wanted to mention it. 

(Github won't let me upload the video here but I added it as a comment to this PR in Slack: https://mitrecorp.slack.com/archives/CLH34KL6A/p1589232347007300?thread_ts=1589058599.028800&cid=CLH34KL6A)

src/components/Builder/Builder.tsx
```diff
@@ -0,0 +1,52 @@
...
+import Header from 'components/Header';
+import Navigation from 'components/Navigation';
+import Sidebar from 'components/Sidebar';
+import Graph from 'components/Graph';
+import Loading from 'components/elements/Loading';
+
+import styles from './Builder.module.scss';
+
+interface BuilderProps {
+  pathway?: Pathway | null;
```
blangley28, line 13: I don't think should be optional anymore. What is the point in viewing the builder screen without a pathway? It isn't even accessible unless a pathway is provided anyway
```suggestion
  pathway: Pathway;
```

src/components/Builder/Builder.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+        {pathway ? (
+          <div ref={graphContainerElement} className={styles.graph}>
+            <Graph pathway={pathway} expandCurrentNode={true} />
+          </div>
+        ) : (
+          <div className={styles.graph}>
+            <Loading />
+          </div>
+        )}
```
blangley28, line 46: Making the pathway required simplifies this
```suggestion
          <div ref={graphContainerElement} className={styles.graph}>
            <Graph pathway={pathway} expandCurrentNode={true} />
          </div>
```

src/components/BuilderRoute.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const pathway = useMemo(() => pathways.find(pathway => pathway.id === pathwayId), [
+    pathwayId,
+    pathways
+  ]);
+  const currentNode = pathway?.states?.[decodeURIComponent(nodeId)];
+
+  if (pathway != null && currentNode == null) {
+    return <Redirect to={`/builder/${id}/node/Start`} />;
+  }
+
```
blangley28, line 20: Add logic here for if the pathway could not be found

src/components/Header/__tests__/Header.test.tsx
```diff
@@ -1,10 +1,15 @@
...
 import { render } from '@testing-library/react';
+import { MemoryRouter } from 'react-router-dom';
 import Header from '../Header';
 
 describe('<Header />', () => {
   it('renders a visible header title and logo', () => {
-    const { getByRole } = render(<Header logo="logo" />);
+    const { getByRole } = render(
+      <MemoryRouter>
+        <Header logo="logo" />
```
blangley28, line 10: This prop was updated to load the logo itself instead of being passed a prop
```suggestion
        <Header />
```

src/components/PathwaysList/NewPathwayModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+  Dialog,
+  DialogTitle,
+  DialogContent,
+  DialogActions,
+  TextField,
+  IconButton
+} from '@material-ui/core';
+import { v4 as uuidv4 } from 'uuid';
+
+import { usePathwayContext } from '../PathwayProvider';
```
blangley28, line 15: I'm assuming we're using the same style guide for the builder. This should be an absolute import

src/components/PathwaysList/NewPathwayModal.tsx
```diff
@@ -0,0 +1,93 @@
...
+import {
+  Button,
+  Dialog,
+  DialogTitle,
+  DialogContent,
+  DialogActions,
+  TextField,
+  IconButton
+} from '@material-ui/core';
+import { v4 as uuidv4 } from 'uuid';
```
blangley28, line 13: This is super minor but I know in other places we use shortid to generate ids. I feel like we should only use one just to be consistent and so we are not importing multiple things to do the same job

src/components/PathwaysList/PathwaysList.tsx
```diff
@@ -1,157 +1,27 @@
...
-  if (service.status === 'loading') return <Loading />;
+  const { status } = usePathwayContext();
 
-  const handleClose = (): void => {
-    // closes modal and resets new pathway state
-    setOpen(false);
-    setNewPathwayName('');
-    setNewPathwayDescription('');
-  };
+  const handleClickOpen = useCallback((): void => {
```
blangley28, line 18: Suggest renaming this to something like `showCreatePathwayModal`

src/components/PathwaysList/PathwaysList.tsx
```diff
@@ -1,157 +1,27 @@
...
-          label: 'Start',
-          transitions: []
-        }
-      }
-    };
-    pathways.push(newPathway);
-    handleClose();
-    callback(newPathway);
-  };
+  const handleClose = useCallback((): void => {
```
blangley28, line 22: Suggest renaming this to something like `hideCreatePathwayModal`

src/components/PathwaysList/PathwaysTable.tsx
```diff
@@ -0,0 +1,79 @@
...
+  Table,
+  TableBody,
+  TableCell,
+  TableContainer,
+  TableHead,
+  TableRow
+} from '@material-ui/core';
+
+import { Pathway } from 'pathways-model';
+import { usePathwayContext } from '../PathwayProvider';
```
blangley28, line 16: Absolute import here too

src/components/PathwaysList/PathwaysTable.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+import useStyles from './styles';
+
+const PathwaysTable: FC = () => {
+  const styles = useStyles();
+  const { pathways } = usePathwayContext();
+
+  function deletePathway(pathway: Pathway | null): void {
+    // TODO
+  }
```
blangley28, line 26: Since this doesn't maintain a list of all the pathways this should probably be moved to the PathwayProvider

src/components/Navigation/Navigation.tsx
```diff
@@ -1,41 +1,31 @@
...
-  }),
-  { name: 'Navigation' }
-);
+import { Pathway } from 'pathways-model';
+import useStyles from './styles';
 
 interface Props {
-  name?: string;
-  onBack?: () => void;
+  pathway?: Pathway | null;
```
blangley28, line 11: Since this was moved to be part of the Builder View only this should be required (assuming you agree with my previous comment about pathway being required for the Builder view)

src/components/PathwaysService/PathwaysService.tsx
```diff
@@ -15,14 +15,21 @@ function typedFetch<T>(url: string, options?: object): Promise<T> {
...
 
   useEffect(() => {
     getPathways(url)
-      .then(response => response.json() as Promise<Array<string>>)
-      .then(listOfFiles => listOfFiles.map(f => typedFetch<Pathway>(url + '/' + f)))
+      .then(response => response.json() as Promise<string[]>)
+      .then(listOfFiles =>
+        listOfFiles.map(f =>
+          typedFetch<Pathway>(url + '/' + f).then(pathway => ({
+            id: f,
```
blangley28, line 28: Instead of adding the id on as an after thought we should probably just add it to the pathway files. When we export the pathway it strips the ids and keys added but we will probably want to save the pathway with this, especially since pathways could probably be loaded from multiple directories meaning filenames are not unique

src/components/PathwaysList/NewPathwayModal.tsx
noranda: Oops I didn't see we already had an id generator. Fixed!

src/components/PathwaysList/NewPathwayModal.tsx
noranda: You are correct! Fixed.

src/components/Graph/Graph.tsx
noranda: Pushed up another fix. Hopefully this now has the same behavior as master.

src/types/pathways-model.d.ts
```diff
@@ -52,9 +55,9 @@ declare module 'pathways-model' {
   }
 
   export interface PathwayContextInterface {
```
danlee1025, line 57: Is it better to have this in the `PathwayProvider`? It seems the other providers that we have define the context interface in the file.

noranda: Welp, I tried moving it and Typescript was not happy so I reverted it back. It makes no sense to me, but if you can figure out how to move it without Typescript making a stink, let me know and I'll move it. ;)

danlee1025: Wow very weird...

Seems you have to give the parameters a name
```
  interface PathwayContextInterface {
    pathways: Pathway[];
    status: string;
    addPathway: (pathway: Pathway) => void;
    deletePathway: (pathway: Pathway) => void;
  }
```

I have no idea why Typescript doesn't complain about it in the definitions file.

noranda: Fixed!

PR #8, opened 2020-05-11
Update infrastructure

blangley28:
Implement functionality for creating a pathway. Created methods to add, update, and delete all of the pathway components plus unit tests for each function. Required modifying the model slightly since objects required a unique key to access them for editing.

### DIFF
--- a/public/static/pathways/her2_pathway.json
+++ b/public/static/pathways/her2_pathway.json
@@ -5,11 +5,13 @@
   "library": "mCODE_Library.cql",
   "criteria": [
     {
+      "id": "1",
       "elementName": "Condition",
       "expected": "Breast Cancer",
       "cql": "\"Primary Cancer Condition\" PrimaryCancer return Tuple{ value: PrimaryCancer.code.text.value, match: PrimaryCancer.code.text.value ~ 'Malignant neoplasm of breast (disorder)' } "
     },
     {
+      "id": "2",
       "elementName": "HER2 Status",
       "expected": "Positive",
       "cql": "[Observation: \"HER2 [Presence] in Breast cancer specimen by Immune stain\"] Her2 return Tuple{ value: Her2.value.text.value, match: Her2.value.text.value ~ 'Positive' } "
@@ -17,17 +19,21 @@
   ],
   "states": {
     "Start": {
+      "key": "Start",
       "label": "Start",
       "transitions": [
         {
+          "id": "1",
           "transition": "NodeStatus"
         }
       ]
     },
     "NodeStatus": {
+      "key": "NodeStatus",
       "label": "Node Status",
       "transitions": [
         {
+          "id": "1",
           "transition": "TumorSize",
           "condition": {
             "description": "N0",
@@ -35,6 +41,7 @@
           }
         },
         {
+          "id": "2",
           "transition": "ChemotherapyTCHAC+TH",
           "condition": {
             "description": "N+",
@@ -44,9 +51,11 @@
       ]
     },
     "TumorSize": {
+      "key": "TumorSize",
       "label": "Tumor Size",
       "transitions": [
         {
+          "id": "1",
           "transition": "ChemotherapyTCHAC+TH",
           "condition": {
             "description": "T > 2cm",
@@ -54,6 +63,7 @@
           }
         },
         {
+          "id": "2",
           "transition": "ChemotherapyTHWeekly",
           "condition": {
             "description": "T <= 2cm",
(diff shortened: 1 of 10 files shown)


src/utils/builder.ts
```diff
@@ -0,0 +1,297 @@
...
+  stateKey: string,
+  actionId: string,
+  resource: MedicationRequest | ServiceRequest
+): void {
+  if (pathway.states[stateKey].action) {
+    const action = pathway.states[stateKey].action.find((action: Action) => action.id === actionId);
+    action.resource = resource;
+  }
+}
+
```
mdube7, line 239: We need a way to switch a branch node to an action node and an action node to a branch node.

src/utils/builder.ts
```diff
@@ -0,0 +1,297 @@
...
+  });
+
+  return JSON.stringify(pathway, undefined, 2);
+}
+
+export function addLibrary(pathway: Pathway, library: string): void {
+  pathway.library = library;
+}
+
+// TODO: possibly add more criteria methods
```
mdube7, line 46: We need Criteria Source to be added to the model

src/utils/builder.ts
blangley28: Added

public/static/pathways/her2_pathway.json
```diff
@@ -4,36 +4,43 @@
   "library": "mCODE_Library.cql",
   "criteria": [
     {
+      "id": "1",
```
dehall, line 7: Do we need an id field for everything in an array? Is it not enough to just use indexes? That would make things simpler in a few places.

public/static/pathways/her2_pathway.json
```diff
@@ -4,36 +4,43 @@
...
       "elementName": "Condition",
       "expected": "Breast Cancer",
       "cql": "\"Primary Cancer Condition\" PrimaryCancer return Tuple{ value: PrimaryCancer.code.text.value, match: PrimaryCancer.code.text.value ~ 'Malignant neoplasm of breast (disorder)' } "
     },
     {
+      "id": "2",
```
dehall, line 13: Also based on the `exportPathway` function below I thought the idea was that these would be deleted in the final exported version

src/utils/builder.ts
```diff
@@ -0,0 +1,310 @@
...
+    elementName: elementName,
+    expected: expected,
+    cql: cql
+  };
+  pathway.criteria.push(criteria);
+
+  return id;
+}
+
+export function addNavigationalElm(pathway: Pathway, elm: object): void {
```
dehall, line 65: Minor point but we should standardize on naming convention. Generally `addX` means adding to a list and `setX` sets a value so this and a few other places should be `setX`

public/static/pathways/her2_pathway.json
blangley28: No, because the id needs to be a way to always identify the object in the list. If we have 3 criteria and delete the middle one, then the identifier of the third criteria changes. I suppose we could make the front end remember the list ordering but I wanted to make the infrastructure versatile.

public/static/pathways/her2_pathway.json
blangley28: Yes it is deleted in the exported version

src/utils/builder.ts
blangley28: Update so only methods adding to a list are called `addX` and those which are updating a value to be `setX`

src/utils/builder.ts
blangley28: From our discussion I don't think is required anymore. The Criteria source is only used to determine what the CQL is. We then use the setCQL method to set the actual CQL. Our model currently only supports libraries at the pathways level and there is a method to set that.

public/static/pathways/her2_pathway.json
dehall: I guess my question was, why do we need to add them in this file then?

blangley28: Ohh, I didn't think was the exported version. I thought these were like the auto saved files. I suppose they can be removed here and added automatically when the pathway is loaded but that seems unnecessary to save the files in a different format from how we need to build them.

dehall: I was assuming we want the same backend to serve up pathways both to Camino and the builder. There are a couple different options if we do that. We could keep them in the final saved version and just ignore them in Camino. We could delete them on export and re-add them on import. 

If we don't want the same backend to host pathways for both apps, then we can save in one format for the builder then export without the temporary keys.

There are probably more options too. None are perfect so I'm open to suggestions.

PR #9, opened 2020-05-12
Add branch node

mdube7:
Putting this up so everyone can start taking a look at it. 

1. Select a Pathway
2. Change the Node Type from Action to Branch or Click the Add Branch Node Button
3. Observe the Sidebar as you fill it out

As a note this task will not be implementing the adding of the nodes to the pathway, but is focused on the design of the sidebar for doing so.

Selecting duplicated node names, like Surgery in the Neoadjuvant pathway, and trying to branch will not work but should not be a problem once we implement PATHWAYS-256.


src/components/Sidebar/AddBranchNode.tsx
```diff
@@ -0,0 +1,74 @@
+import React, { FC } from 'react';
```
blangley28, line 1: Add useState and ChangeEvent as an import here 
```suggestion
import React, { FC, useState, ChangeEvent } from 'react';
```

src/components/Sidebar/AddBranchNode.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+interface BranchNodeProps {
+  currentNode: State;
+  addChoiceNode: (event: React.MouseEvent<HTMLButtonElement>) => void;
+}
+
+const AddBranchNode: FC<BranchNodeProps> = ({ currentNode, addChoiceNode }) => {
+  // TODO: in PATHWAYS-256 use the properties in the pathway instead?
+  const [source, setSource] = React.useState('');
+  const [critera, setCritera] = React.useState('');
```
blangley28, line 18: And then remove `React` here. Also for typescript it should include the type parameter (looks like both of these are used as strings below)
```suggestion
  const [source, setSource] = useState<string>('');
  const [critera, setCritera] = useState<string>('');
```

src/components/Sidebar/AddBranchNode.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  const handleSourceChange = (event: React.ChangeEvent<{ value: unknown }>): void => {
```
blangley28, line 20: And then remove React here
```suggestion
  const handleSourceChange = (event: ChangeEvent<{ value: unknown }>): void => {
```

src/components/Sidebar/AddBranchNode.tsx
```diff
@@ -0,0 +1,74 @@
...
+    // TODO: in PATHWAYS-256 set the source
+    setSource(event.target.value as string);
+    setCritera('');
+  };
+
+  const handleCriteraChange = (event: React.ChangeEvent<{ value: unknown }>): void => {
```
blangley28, line 26: Also remove React here
```suggestion
  const handleCriteraChange = (event: ChangeEvent<{ value: unknown }>): void => {
```

src/components/Sidebar/AddChoiceButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { Button } from '@material-ui/core';
+
+interface AddChoiceButtonProps {
+  addChoiceNode: (event: React.MouseEvent<HTMLButtonElement>) => void;
+}
+
+// TODO: in PATHWAYS-256 Add choice needs to update the pathway with the new transition and state
+const AddChoiceButton: FC<AddChoiceButtonProps> = ({ addChoiceNode }) => {
+  return (
+    <table>
```
blangley28, line 14: In other places we use the Material UI Table, should we be using that here as well for consistency?

src/components/Sidebar/AddChoiceButton.tsx
```diff
@@ -0,0 +1,36 @@
...
+      <tbody>
+        <tr>
+          <td className={styles.button}>
+            <Button
+              variant="contained"
+              color="primary"
+              startIcon={<FontAwesomeIcon icon={faPlus} />}
+              onClick={addChoiceNode}
+            >
+              Add Choice Node
```
blangley28, line 24: I know "Choice" is from the design but how do people think about renaming it to "Transition"

src/components/Sidebar/ChoiceNode.tsx
```diff
@@ -0,0 +1,83 @@
+import React, { FC, ChangeEvent } from 'react';
```
blangley28, line 1: Add useState here 
```suggestion
import React, { FC, ChangeEvent, useState } from 'react';
```

src/components/Sidebar/ChoiceNode.tsx
```diff
@@ -0,0 +1,83 @@
...
+      flexWrap: 'wrap'
+    }
+  }));
+
+  const classes = useStyles();
+  const initial = pathway !== null ? pathway.states[transition].label : 'Choice Node';
+
+  // TODO: in PATHWAYS-256 use the data from the pathway
+  const [choice, setChoice] = React.useState('');
+  const [label, setLabel] = React.useState<string>(initial);
```
blangley28, line 35: Remove React here and add type to choice
```suggestion
  const [choice, setChoice] = useState<string>('');
  const [label, setLabel] = useState<string>(initial);
```

src/components/Sidebar/Sidebar.tsx
```diff
@@ -109,15 +173,23 @@ const Sidebar: FC<SidebarProps> = ({ headerElement }) => {
...
+            >
+              <MenuItem value={'action'}>Action</MenuItem>
+              <MenuItem value={'branch'}>Branch</MenuItem>
+            </Select>
+          </FormControl>
+          {currentNode.nodeType === 'branch' ? (
+            <AddBranchNode currentNode={currentNode} addChoiceNode={addChoiceNoide} />
+          ) : (
+            <AddNodes addBranchNode={addBranchNode} />
+          )}
```
blangley28, line 192: I think the naming is a little confusing here. If the current node is branch then we can add choice (transition nodes) but otherwise we show a button to add a new branch node right? But the component name in the first case is `AddBranchNode`

src/types/pathways-model.d.ts
```diff
@@ -26,6 +26,8 @@ declare module 'pathways-model' {
   export interface State {
     label: string;
     transitions: Transition[];
+    // TODO: in PATHWAYS-256 remove this property
+    nodeType?: string;
```
blangley28, line 30: I don't think this needs to wait for PATHWAYS-256 to be pulled out does it?

src/components/Sidebar/ChoiceNode.tsx
```diff
@@ -0,0 +1,83 @@
...
+    // and the state of the choice node
+    setLabel(event.target.value as string);
+  };
+
+  // TODO: in PATHWAYS-256 the forward button needs to select the newly created choice node
+  return (
+    <div>
+      <div className={styles.choiceNode}>
+        <FormControl className={classes.formControl}>
+          <Input
```
blangley28, line 53: This is a design question for @noranda should the name be editable by simply clicking on it? Or should the user need to click the edit icon first?

noranda: The entire form element should be clickable to edit. The edit icon is a visual indicator that the field is editable, so it's more of an [input adornment](https://material-ui.com/components/text-fields/#input-adornments).

src/types/pathways-model.d.ts
mdube7: I am using this property in sidebar to determine which "type" of sidebar to display

src/components/Sidebar/Sidebar.tsx
mdube7: If we renamed the component to `BranchNode` would that make it more clear?

src/components/Sidebar/AddChoiceButton.tsx
noranda: Yes.

src/components/Sidebar/AddChoiceButton.tsx
noranda: I like it.

src/components/Sidebar/McodeCriteria.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { FC, ChangeEvent, memo } from 'react';
```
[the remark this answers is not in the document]
mdube7: combined the two into one component

src/components/Sidebar/AddChoiceButton.tsx
noranda: Actually, I don't think a table is even needed here. We can get by using divs and flexbox.

src/components/Sidebar/Sidebar.tsx
noranda: Renamed to `BranchNode`.

noranda:
With Mike's permission, I've made a few commits to do the following:

1. Sidebar components were restructured and renamed to be more readable.
2. Sidebar functions were refactored to use the new builder utils added in the last PR. As a note, without an API, the logic for updating the data models are complex. I recommend we revisit the need for an API sooner than later.
3. The graph will now update when you add or edit node details.
4. You can now edit node names, which will update in real time in the graph.
5. Styles were refactored and updated.

There is still a lot of work to do to get the graph to work well with building up nodes. For example, while the new nodes will render, the arrows do not. Recommend adding a JIRA task to address this.

blangley28:
@noranda I will go through the changes again but I have two questions/comments on your last comment. The "builder utils" was supposed to be a sort of API to buid/edit/update the underlying pathway. What do you mean by we need an API?

Also it seems like you started some of the work on [PATHWAYS-256](https://jira.mitre.org/browse/PATHWAYS-256) which deals with linking the sidebar builder to these utils. I can add a comment to this task referencing your work and this task can then handle the work such as rendering the arrows.

My initial intuition is that the arrows aren't rendering because we don't have the UI set up yet to add transitions. I suppose if we are on a selected state and click "Add Node" it should automatically set the transition (from the current node to the new node).

noranda:
@blangley28 By build an API, I just mean if we're going to build a backend, these functions should live there, so there is starting to be a lot of code that is being written that will have to be rewritten if/when we move everything to a backend. This works for now though. Regarding PATHWAYS-256, I now realize I've implemented a good portion of that task, but that was also done to reduce having to rewrite code at a later time. There are still issues with the hook up that will need to be addressed, like the arrows not rendering for example. Thanks for taking a look!

src/components/Sidebar/BranchTransition.tsx
```diff
@@ -0,0 +1,41 @@
...
+      <SidebarButton
+        buttonName="Use Criteria"
+        buttonIcon={<FontAwesomeIcon icon={faPlus} />}
+        buttonText="Add previously built or imported criteria logic to branch node."
+      />
+
+      <SidebarButton
+        buttonName="Build Criteria"
+        buttonIcon={<FontAwesomeIcon icon={faTools} />}
+        buttonText="Create new criteria logic to add to branch node.."
```
blangley28, line 35: Looks like an extra period
```suggestion
        buttonText="Create new criteria logic to add to branch node."
```

src/types/pathways-model.d.ts
```diff
@@ -30,6 +30,7 @@ declare module 'pathways-model' {
     key?: string;
     label: string;
     transitions: Transition[];
+    nodeType?: string;
```
blangley28, line 33: I still think we should remove this

src/utils/builder.ts
```diff
@@ -69,46 +69,131 @@ export function setCriteriaElm(pathway: Pathway, elm: object): void {
...
+export function addGuidanceState(pathway: Pathway): Pathway {
+  const state = createState();
+  const newPathway = addState(pathway, state);
+
+  return makeBranchStateGuidance(newPathway, state.key as string);
 }
 
-export function setStateLabel(pathway: Pathway, key: string, label: string): void {
-  pathway.states[key].label = label;
+export function setStateLabel(pathway: Pathway, key: string, label: string): Pathway {
```
blangley28, line 100: Is there a reason this changed?

src/utils/builder.ts
```diff
@@ -69,46 +69,131 @@ export function setCriteriaElm(pathway: Pathway, elm: object): void {
   pathway.elm.criteria = elm;
 }
 
-function addState(pathway: Pathway, key?: string): string {
+export function createState(key?: string): State {
```
blangley28, line 72: Why does this need to be exported if the branch state and guidance state methods are exported?

src/utils/builder.ts
```diff
@@ -69,46 +69,131 @@ export function setCriteriaElm(pathway: Pathway, elm: object): void {
...
+  key: string,
+  criteriaSource: string
+): Pathway {
+  return {
+    ...pathway,
+    states: {
+      ...pathway.states,
+      [key]: {
+        ...pathway.states[key],
+        criteriaSource
```
blangley28, line 137: CriteriaSource was originally left off the interface because the Camino app does not support it

src/components/Sidebar/Sidebar.tsx
```diff
@@ -1,115 +1,62 @@
...
-              </Button>
-            </td>
-            <td className={styles.description}>
-              A previously built node or group of nodes defining a set of criteria.
-            </td>
-          </tr>
-        </tbody>
-      </table>
-    </div>
+  const changeNodeType = useCallback(
```
blangley28, line 29: Unless I am missing something it looks like all this is doing is changing the nodeType indicator and not actually changing the node type. The methods `makeBranchStateGuidance` and `makeGuidanceStateBranch` handle this

src/components/Sidebar/BranchTransition.tsx
noranda: Whoops!

src/utils/builder.ts
noranda: Yes. The previous implementation was modifying the pathway in place, but React doesn't know how to handle that. React only knows to re-render when it sees a new object. That's why in Redux we create new objects in actions, and that's what we have to do here.

src/utils/builder.ts
noranda: We need to export this to be used in the `BranchNode` component.

src/utils/builder.ts
noranda: Yeah this will go away with the v2 designs. I left it in because it was in the v1 design and implemented with this task.

blangley28: V2 being [PATHWAYS-259](https://jira.mitre.org/browse/PATHWAYS-259)?

src/utils/builder.ts
blangley28: Why can't it just use the `addBranchState` function?

src/utils/builder.ts
blangley28: That's really frustrating...

src/components/Sidebar/Sidebar.tsx
noranda: I believe this is addressed in the latest commits.

src/utils/builder.ts
noranda: This was reworked to simplify the function. It will return a basic state object, which is then added to a pathway in `addNode()` in the sidebar component. This then lets us do things like add transitions and set the node type.

src/types/pathways-model.d.ts
```diff
@@ -30,17 +30,19 @@ declare module 'pathways-model' {
...
   }
 
-  // NOTE: the model also includes a BranchState (which extends State),
-  // but as of right now it has no additional fields not in State,
-  // and TypeScript does not allow "empty" interfaces so we can't add it yet.
-  // Add it here if/when we ever need it.
+  export interface BranchState extends State {
+    criteriaSource?: string;
+    mcodeCriteria?: string;
+    otherCriteria?: string;
```
blangley28, line 44: Do we need both of these? Isn't it implied whether or not it is mcode criteria based on the source?

src/utils/builder.ts
```diff
@@ -20,24 +20,38 @@ export function createNewPathway(name: string, description?: string, pathwayId?:
 }
 
 export function exportPathway(pathway: Pathway): string {
```
blangley28, line 22: Setting properties to undefined isn't really removing them. The exported pathway will still have the property there.

src/utils/builder.ts
```diff
@@ -231,24 +347,61 @@ export function setActionResource(
...
+
+  return {
+    ...pathway,
+    states: {
+      ...pathway.states,
+      [stateKey]: {
+        ...state,
+        cql: '',
+        action: [],
+        nodeTypeIsUndefined: undefined,
```
blangley28, line 373: Same here making these properties undefined

src/utils/builder.ts
```diff
@@ -231,24 +347,61 @@ export function setActionResource(
...
+    return pathway;
+  }
+
+  return {
+    ...pathway,
+    states: {
+      ...pathway.states,
+      [stateKey]: {
+        ...state,
+        cql: undefined,
```
blangley28, line 399: And here

blangley28:
I didn't have a full time to review in detail and since I'm about to take PTO don't let me hold this up. But I will say it seems like a lot changed in the builder. I know we discussed how React state requires a new object so that had to change but I don't think the tests should be different. The new methods added should have new tests but the point of the interface is that the internal method can be changed without changing the impact of the calling the method. As I said, I didn't have time to take a super quick look and perhaps there was a good need for them to change but it seems like a lot changed.

src/utils/builder.ts
noranda: When the pathway is exported, `JSON.stringify()` removes any properties with an undefined property. Because we need to create new pathways, I find this easier to manage.

src/types/pathways-model.d.ts
noranda: Both of these will be removed in a future task when we implement branch criteria.

noranda:
> I didn't have a full time to review in detail and since I'm about to take PTO don't let me hold this up. But I will say it seems like a lot changed in the builder. I know we discussed how React state requires a new object so that had to change but I don't think the tests should be different. The new methods added should have new tests but the point of the interface is that the internal method can be changed without changing the impact of the calling the method. As I said, I didn't have time to take a super quick look and perhaps there was a good need for them to change but it seems like a lot changed.

@blangley28 You're right, there are a lot of changes here, but I'd argue that they're all necessary. The bulk of the changes to the builder utils involved making sure we don't modify pathways in place, which can create issues with rendering. There was also a bug with the `Pathways` interface because we were using `BranchState`, which wasn't defined and so defaulted to `any`. Fixing that meant fixing all the typescript errors that popped up. The remainder of the changes were minor including fixing some uncontrolled inputs. If you have any questions about anything, just ask!

src/utils/builder.ts
mdube7: See Noranda's comment above

src/utils/builder.ts
mdube7: See Noranda's comment above

src/components/Sidebar/ActionNode.tsx
```diff
@@ -0,0 +1,70 @@
...
+          value="action"
+        />
+      )}
+      {displayAddButtons && (
+        <>
+          {changeNodeTypeEnabled && <hr className={styles.divider} />}
+          <SidebarButton
+            buttonName="Add Action Node"
+            buttonIcon={<FontAwesomeIcon icon={faPlus} />}
+            buttonText="Any clinical or workflow step which is not a decision."
```
dehall, line 54: This text is a little unclear, and we should avoid the word decision because technically an action node does involve a decision to accept or decline. The idea behind an action node is that it's an atomic recommendation for treatment, for example a medication or a procedure.

noranda: Good point. I think it would be helpful to get Jim's input for the text on all the buttons.

PR #10, opened 2020-05-27
Criteria List

dehall:
Adds tabs to the main page, along with a new tab for Criteria. On the Criteria tab we list out the current Criteria maintained by the CriteriaProvider, and the ability to upload ELM to import new criteria. 

This is a combination of PATHWAYS-260 (Add tabs and Criteria list tab to main selection page) and PATHWAYS-239 (Implement Criteria Provider and importing from criteria tab)


dehall:
Note that the tabs use the `react-tabs` library and not Material UI. I experimented with the Material UI tabs for a day but there were some downsides and I didn't want to spend any more time trying to style it. Below are notes on 2 possible approaches in case we do want to go that direction


1. Core Library Only
Downside: we need to do our own state management and conditional rendering of tab content

```
import { Tab, Tabs } from '@material-ui/core';

const [value, setValue] = useState<number>(0);

<Tabs 
  value={value} 
  onChange={(event: object, value: any):void => setValue(value) } >
  <Tab label='Pathways' />
  <Tab label='Criteria' />
</Tabs>

{ value === 0 && <PathwaysList /> }
{ value === 1 && <CriteriaList /> }
```

2. Core plus Labs (beta)
Downside: this is bleeding-edge (merged within the last 2 weeks) and there may be bugs
One notable bug is that as of when I tried it there was a console warning about a `<div>` within a `<p>`
Should be fixed on master (https://github.com/mui-org/material-ui/pull/21017) but is not yet on npm

Still need to do our own state management though TabPanel handles conditional rendering

```
import { Tab } from '@material-ui/core';
import { TabContext, TabList, TabPanel } from '@material-ui/lab';

const [value, setValue] = useState<string>('pathways');

<TabContext value={value}>
  <TabList onChange={(event: object, value: any):void => setValue(value) }>
    <Tab label='Pathways' value='pathways' />
    <Tab label='Criteria' value='criteria' />
  </TabList>

  <TabPanel value='pathways'>
    <PathwaysList />
  </TabPanel>
  <TabPanel value='criteria'>
    <CriteriaList />
  </TabPanel>
</TabContext>
```

src/components/CriteriaProvider.tsx
```diff
@@ -0,0 +1,86 @@
...
+  FC,
+  useState,
+  memo,
+  useContext,
+  useCallback
+} from 'react';
+import shortid from 'shortid';
+import { ElmStatement } from 'pathways-model';
+
+interface Criteria {
```
danlee1025, line 13: I see that there is another `Criteria` interface in `pathways-model.d.ts`... I'm assuming these are different?

blangley28: Yea this is why I opposed the name Criteria originally but nothing better came up. In the sense of this PR all of the "Criteria" is CQL/ELM used for navigating the pathway (transitions and marking nodes as complete). The other "Criteria" which already exists is `pathway.criteria`. This Criteria is for determining the applicability of the pathway

src/components/App.tsx
```diff
@@ -1,33 +1,50 @@
...
+              <Switch>
+                <Route path="/builder/:id/node/:nodeId">
+                  <BuilderRoute />
+                </Route>
+                <Route path="/builder/:id">
+                  <BuilderRoute />
+                </Route>
+                <Route path="/">
+                  <Header />
+                  <Tabs>
```
radamson, line 31: It looks like the Tabs here could be broken out into their own component to help keep `App.tsx` manageable/readable while maintain separation of concerns and all that other good stuff.

It's not much, but in the long run it might save us some headache/refactoring.  Even if the other component is still defined in this same file.

I'm thinking about this block of code specifically:

```jsx
                  <Tabs>
                    <TabList>
                      <Tab>Pathways</Tab>
                      <Tab>Criteria</Tab>
                    </TabList>

                    <TabPanel>
                      <PathwaysList />
                    </TabPanel>
                    <TabPanel>
                      <CriteriaList />
                    </TabPanel>
                  </Tabs>
```

src/components/CriteriaList/CriteriaList.tsx
```diff
@@ -0,0 +1,55 @@
...
+import CriteriaTable from './CriteriaTable';
+import ImportCriteriaModal from './ImportCriteriaModal';
+
+import useStyles from './styles';
+
+const CriteriaList: FC = () => {
+  const styles = useStyles();
+  const { status } = usePathwayContext();
+
+  const [open, setOpen] = useState(false);
```
radamson, line 17: Set the type

```suggestion
  const [open, setOpen] = useState<boolean>(false);
```

src/components/CriteriaList/ImportCriteriaModal.tsx
```diff
@@ -0,0 +1,73 @@
...
+const ImportCriteriaModal: FC<ImportCriteriaModalProps> = ({ open, onClose }) => {
+  const styles = useStyles();
+  const importFileRef = useRef<HTMLInputElement>(null);
+
+  const { addCriteria } = useCriteriaContext();
+
+  const selectFile = useCallback(
+    (event: FormEvent<HTMLFormElement>) => {
+      event.preventDefault();
+      const files = importFileRef.current?.files;
```
radamson, line 31: it looks like `importedFileRef` defaults to null, so it might be prudent to have an additional optional chaining operator here

```suggestion
      const files = importFileRef?.current?.files;
```

src/styles/_tabs.scss
```diff
@@ -0,0 +1,57 @@
...
+  border-bottom: none;
+  bottom: -1px;
+  position: relative;
+  list-style: none;
+  padding: 6px 12px;
+  cursor: pointer;
+}
+
+.react-tabs__tab--selected {
+  background: #fff;
```
radamson, line 24: `#fff` is already in `_variables.scss` as `$white`

```suggestion
  background: $white;
```

src/styles/_tabs.scss
```diff
@@ -0,0 +1,57 @@
...
+  border-color: #aaa;
+  color: black;
```
radamson, line 26: We should be using the color from our palette to make theming and changing colors easier.

```suggestion
  color: $black;
```

src/styles/_tabs.scss
```diff
@@ -0,0 +1,57 @@
...
+  bottom: -1px;
+  position: relative;
+  list-style: none;
+  padding: 6px 12px;
+  cursor: pointer;
+}
+
+.react-tabs__tab--selected {
+  background: #fff;
+  border-color: #aaa;
```
radamson, line 25: I didn't see this color specifically in the variables file.  Is there another one we should be using or should this be added to the palette in `_variables.scss`?

src/styles/_tabs.scss
```diff
@@ -0,0 +1,57 @@
...
+}
+
+.react-tabs__tab:focus:after {
+  content: "";
+  position: absolute;
+  height: 5px;
+  left: -4px;
+  right: -4px;
+  bottom: -5px;
+  background: #fff;
```
radamson, line 48:
```suggestion
  background: $white;
```

src/styles/_tabs.scss
```diff
@@ -0,0 +1,57 @@
...
+}
+
+.react-tabs__tab--disabled {
+  color: GrayText;
+  cursor: default;
+}
+
+.react-tabs__tab:focus {
+  box-shadow: 0 0 5px hsl(208, 99%, 50%);
+  border-color: hsl(208, 99%, 50%);
```
radamson, line 37: Same with this color.  it is repeated here so it probably belongs with the other color variables.

src/utils/builder.ts
```diff
@@ -20,26 +29,131 @@ export function createNewPathway(name: string, description?: string, pathwayId?:
...
+            }
+          }
+        ]
+      },
+      includes: { def: [] },
+      valueSets: { def: [] }
+    }
+  };
+
+  const mergeElm = (additionalElm: ElmLibrary): void => {
```
radamson, line 79: I think it would be better to define this as another function alongside its encapsulating method `exportPathway` and just pass `elm` in as an argument.

Although the closure is nice this method is going to be redefined everytime it runs, `exportPathway` is already >100 lines long, and it only has a single argument as is.  Plus, I could see this method being useful elsewhere in the future.

src/styles/_tabs.scss
```diff
@@ -0,0 +1,57 @@
...
+  background: #fff;
+}
+
+.react-tabs__tab-panel {
+  display: none;
+}
+
+.react-tabs__tab-panel--selected {
+  display: block;
+}
```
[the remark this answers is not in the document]
noranda: Note: This comment makes more sense after reading the next comment. ;)

src/styles/_tabs.scss
blangley28: Agreed. This is part of Dylan's work and I'm not actually sure what it is styling. Could you tell me what color it is and I will make it a variable?

src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+const Tabs: FC<TabsProp> = ({ tabs }) => {
+  const styles = useStyles();
+
+  return (
+    <div className={styles.root}>
+      <ReactTabs>
+        <TabList>
+          {tabs.map(tab => (
+            <ReactTab>
```
noranda, line 25: Need to add a key here (I missed this earlier).
```js
{tabs.map((tab, index) => (
  <ReactTab key={index}>
    ...
```

src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,39 @@
...
+              <Button classes={{ label: styles.label }}>{tab.label}</Button>
+            </ReactTab>
+          ))}
+        </TabList>
+
+        {tabs.map(tab => (
+          <TabPanel>{tab.component}</TabPanel>
```
noranda, line 32: Also need a key here.

```js
{tabs.map((tab, index) => (
  <TabPanel key={index}>{tab.component}</TabPanel>
    ...
```

src/components/Tabs/Tabs.tsx
blangley28: Using the index is not recommended so I used `shortid.generate()` to create a key for these components

src/components/Tabs/Tabs.tsx
```diff
@@ -22,14 +23,14 @@ const Tabs: FC<TabsProp> = ({ tabs }) => {
       <ReactTabs>
         <TabList>
           {tabs.map(tab => (
-            <ReactTab>
+            <ReactTab key={shortid.generate()}>
```
noranda, line 26: While this technically works, it's inefficient because it will generate a different id on every re-render, which means React will destroy this block, even when the contents are exactly the same. I suggested using the index only because we're using a hard-coded array to map over. However, I'm not that concerned about our performance here, so this works. 👍

PR #12, opened 2020-06-08
Update branch node design

mdube7:
Update the branch node to the new design. The `Use Criteria` button uses harder coded values and the  `Build Criteria` button is disabled for now.

### DIFF
--- a/src/components/Sidebar/BranchTransition.tsx
+++ b/src/components/Sidebar/BranchTransition.tsx
@@ -23,7 +23,7 @@ const BranchTransition: FC<BranchTransitionProps> = ({
   updatePathway
 }) => {
   const { criteria } = useCriteriaContext();
-  const criteriaOptions = criteria.map(c => ({ value: c.label, label: c.label }));
+  const criteriaOptions = criteria.map(c => ({ value: c.id, label: c.label }));
   const styles = useStyles();
   const transitionKey = transition?.transition || '';
   const transitionNode = pathway.states[transitionKey];
@@ -40,7 +40,7 @@ const BranchTransition: FC<BranchTransitionProps> = ({
       const criteriaSource = event?.target.value || '';
       let elm = undefined;
       criteria.forEach(c => {
-        if (c.label === criteriaSource) {
+        if (c.id === criteriaSource) {
           elm = c.elm;
         }
       });
@@ -104,10 +104,7 @@ const BranchTransition: FC<BranchTransitionProps> = ({
               value={transition.condition?.description || ''}
               variant="outlined"
               onChange={setCriteriaDisplay}
-              error={
-                transition.condition?.description === undefined ||
-                transition.condition?.description === ''
-              }
+              error={!transition.condition?.description}
             />
           </FormControl>
         </>
--- a/src/components/Sidebar/BranchTransition.tsx
+++ b/src/components/Sidebar/BranchTransition.tsx
@@ -70,6 +70,8 @@ const BranchTransition: FC<BranchTransitionProps> = ({
         setTransitionCriteriaDisplay(pathway, criteriaDisplay || '', transitionKey, currentNodeKey)
       );
     },
+    // Don't include criteriaDisplay as we only want to save when something else is updated
+    // eslint-disable-next-line
     [transition.id, transitionKey, currentNodeKey, updatePathway, pathway]
   );
 
--- a/src/components/Sidebar/BranchTransition.tsx
+++ b/src/components/Sidebar/BranchTransition.tsx
@@ -65,15 +65,9 @@ const BranchTransition: FC<BranchTransitionProps> = ({
 
       const criteriaSource = event?.target.value || '';
       updatePathway(setTransitionCriteria(pathway, criteriaSource, transitionKey, currentNodeKey));
-
       // Need to also save the criteria display as the component will reload without saving
       updatePathway(
-        setTransitionCriteriaDisplay(
-          pathway,
-          savedCriteriaDisplay.current || '',
-          transitionKey,
-          currentNodeKey
-        )
+        setTransitionCriteriaDisplay(pathway, criteriaDisplay || '', transitionKey, currentNodeKey)
       );
     },
     [transition.id, transitionKey, currentNodeKey, updatePathway, pathway]
--- a/src/components/Sidebar/BranchTransition.tsx
+++ b/src/components/Sidebar/BranchTransition.tsx
@@ -3,24 +3,15 @@ import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faPlus, faTools } from '@fortawesome/free-solid-svg-icons';
 import DropDown from 'components/elements/DropDown';
 import { TextField, FormControl } from '@material-ui/core';
-import {
-  setStateCriteriaSource,
-  setStateMcodeCriteria,
-  setStateOtherCriteria,
-  addTransition,
-  createState,
-  addState,
-  setTransitionCriteria,
-  setTransitionCriteriaDisplay
-} from 'utils/builder';
+import { setTransitionCriteria, setTransitionCriteriaDisplay } from 'utils/builder';
 import { SidebarHeader, SidebarButton } from '.';
-import { Pathway, Transition, BranchState } from 'pathways-model';
+import { Pathway, Transition } from 'pathways-model';
 import useStyles from './styles';
 
 interface BranchTransitionProps {
   pathway: Pathway;
   transition: Transition;
-  currentNode: BranchState;
+  currentNodeKey: string;
   updatePathway: (pathway: Pathway) => void;
 }
 
@@ -33,18 +24,17 @@ const criteriaOptions = [
 const BranchTransition: FC<BranchTransitionProps> = ({
   pathway,
   transition,
-  currentNode,
+  currentNodeKey,
   updatePathway
 }) => {
   const styles = useStyles();
   const transitionKey = transition?.transition || '';
   const transitionNode = pathway.states[transitionKey];
   const [useSelected, setUseSelected] = useState<boolean>(false);
-  const currentNodeKey = currentNode?.key;
 
   const handleUseCriteria = useCallback((): void => {
     setUseSelected(!useSelected);
-  }, [pathway, updatePathway]);
+  }, [useSelected]);
 
   const selectCriteriaSource = useCallback(
     (event: ChangeEvent<{ value: string }>): void => {
@@ -53,7 +43,7 @@ const BranchTransition: FC<BranchTransitionProps> = ({
       const criteriaSource = event?.target.value || '';
       updatePathway(setTransitionCriteria(pathway, criteriaSource, transitionKey, currentNodeKey));
     },
-    [currentNodeKey, updatePathway, pathway]
+    [transition.id, transitionKey, currentNodeKey, updatePathway, pathway]
   );
 
   const setCriteriaDisplay = useCallback(
@@ -65,9 +55,9 @@ const BranchTransition: FC<BranchTransitionProps> = ({
         setTransitionCriteriaDisplay(pathway, criteriaSource, transitionKey, currentNodeKey)
       );
     },
-    [currentNodeKey, updatePathway, pathway]
+    [transition.id, transitionKey, currentNodeKey, updatePathway, pathway]
   );
-
+  console.log('transition re-render');
   return (
     <>
       <hr className={styles.divider} />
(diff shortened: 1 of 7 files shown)


src/components/Sidebar/BranchTransition.tsx
```diff
@@ -1,21 +1,119 @@
...
 
 interface BranchTransitionProps {
   pathway: Pathway;
+  currentNodeKey: string;
   transition: Transition;
   updatePathway: (pathway: Pathway) => void;
 }
 
-const BranchTransition: FC<BranchTransitionProps> = ({ pathway, transition, updatePathway }) => {
+const criteriaOptions = [
```
dehall, line 18: Now that you rebased and have the criteria provider to work with, I think it's just a couple lines to integrate it here.

```jsx
import { useCriteriaContext } from 'components/CriteriaProvider';

  const { criteria } = useCriteriaContext();
  const criteriaOptions = criteria.map(c => ({ value: c.label, label: c.label })); // or something else along these lines
```

src/components/Sidebar/BranchTransition.tsx
```diff
@@ -27,12 +125,35 @@ const BranchTransition: FC<BranchTransitionProps> = ({ pathway, transition, upda
...
+          onClick={handleUseCriteria}
+        />
+      )}
 
-      <SidebarButton
-        buttonName="Use Criteria"
-        buttonIcon={<FontAwesomeIcon icon={faPlus} />}
-        buttonText="Add previously built or imported criteria logic to branch node."
-      />
+      {useSelected === true && (
```
blangley28, line 137:
```suggestion
      {useSelected && (
```

src/components/Sidebar/BranchTransition.tsx
```diff
@@ -27,12 +125,35 @@ const BranchTransition: FC<BranchTransitionProps> = ({ pathway, transition, upda
         updatePathway={updatePathway}
         isTransition={true}
       />
+      {useSelected === false && (
```
blangley28, line 128:
```suggestion
      {!useSelected && (
```

src/components/Sidebar/BranchTransition.tsx
```diff
@@ -1,21 +1,119 @@
...
+const BranchTransition: FC<BranchTransitionProps> = ({
+  pathway,
+  currentNodeKey,
+  transition,
+  updatePathway
+}) => {
   const styles = useStyles();
   const transitionKey = transition?.transition || '';
   const transitionNode = pathway.states[transitionKey];
+  const [useSelected, setUseSelected] = useState<boolean>(false);
```
blangley28, line 49: Can we rename this variable to be a little more obvious what it does. It looks like it is used to represented whether or not use criteria button has been selected.

src/types/pathways-model.d.ts
```diff
@@ -39,12 +39,6 @@ declare module 'pathways-model' {
     action: Action[];
   }
 
-  export interface BranchState extends State {
-    criteriaSource?: string;
-    mcodeCriteria?: string;
-    otherCriteria?: string;
-  }
-
```
blangley28, line 47: Since we removed the BranchState interface we should add back the note we had before about it. 

> // NOTE: the model also includes a BranchState (which extends State),
  // but as of right now it has no additional fields not in State,
  // and TypeScript does not allow "empty" interfaces so we can't add it yet.
  // Add it here if/when we ever need it.

src/types/pathways-model.d.ts
```diff
@@ -60,6 +54,8 @@ declare module 'pathways-model' {
       cql: string;
       elm?: ElmLibrary;
     };
+    criteria?: string;
+    criteriaDisplay?: string;
```
blangley28, line 58: The Camino app right now just uses the `Transition.condition.description` for the display. Instead of creating a new property we should just populate that. It saves creating a new method in the builder interface and saves having to strip it when we export the pathway

src/types/pathways-model.d.ts
```diff
@@ -60,6 +54,8 @@ declare module 'pathways-model' {
       cql: string;
       elm?: ElmLibrary;
     };
+    criteria?: string;
```
blangley28, line 57: So I know this isn't linked to the criteria provider yet but we probably want to set the criteria string to be the `transition.condition.cql` and then use that in the future task to populate the `transition.condition.elm`

blangley28: Some of it can be done in Pathways-261 but my main point was we don't need to create a new property in the interface for criteria when one already exists for that use.

mdube7: removed criteria property

mdube7:
I believe this should also include the changes required for 261 now as well.

src/components/Sidebar/BranchTransition.tsx
```diff
@@ -1,22 +1,75 @@
...
 
-const BranchTransition: FC<BranchTransitionProps> = ({ pathway, transition, updatePathway }) => {
+const BranchTransition: FC<BranchTransitionProps> = ({
+  pathway,
+  currentNodeKey,
+  transition,
+  updatePathway
+}) => {
+  const { criteria } = useCriteriaContext();
+  const criteriaOptions = criteria.map(c => ({ value: c.label, label: c.label }));
```
blangley28, line 26: So the label isn't necessarily unique. We should use the id as the value. Then line 43 can be updated to search for the id
```suggestion
  const criteriaOptions = criteria.map(c => ({ value: c.id, label: c.label }));
```

src/components/Sidebar/BranchTransition.tsx
```diff
@@ -1,22 +1,75 @@
...
+    (event: ChangeEvent<{ value: string }>): void => {
+      if (!currentNodeKey || !transition.id) return;
+
+      const criteriaSource = event?.target.value || '';
+      let elm = undefined;
+      criteria.forEach(c => {
+        if (c.label === criteriaSource) {
+          elm = c.elm;
+        }
+      });
```
blangley28, line 46: See comment above about using id
```suggestion
      criteria.forEach(c => {
        if (c.id === criteriaSource) {
          elm = c.elm;
        }
      });
```

src/utils/builder.ts
```diff
@@ -329,25 +275,70 @@ export function setTransitionCondition(
   startNodeKey: string,
   transitionId: string,
   description: string,
-  elm: ElmLibrary
-): void {
+  elm: ElmLibrary,
+  criteria?: string
```
blangley28, line 279: I'm getting confused throwing criteria around - this refers to the criteria label right? Naming it `criteriaLabel` would be clearer since its not the entire criteria object.

src/utils/builder.ts
```diff
@@ -329,25 +275,70 @@ export function setTransitionCondition(
...
 export function setGuidanceStateElm(pathway: Pathway, key: string, elm: ElmLibrary): void {
   (pathway.states[key] as GuidanceState).elm = elm;
   (pathway.states[key] as GuidanceState).cql = getElmStatement(elm).name;
 }
 
+export function setTransitionCql(
+  pathway: Pathway,
+  criteria: string,
+  transitionId: string,
+  startNodeKey: string
```
blangley28, line 314: This is super minor but all of the other methods here follow the format `(pathway, keys/ids, new value)`. Also while we may be passing in the value `criteria.label` when we call the method I think it makes more sense to name the argument `cql` instead of `criteria`

src/components/Sidebar/BranchTransition.tsx
```diff
@@ -27,12 +80,38 @@ const BranchTransition: FC<BranchTransitionProps> = ({ pathway, transition, upda
...
-        buttonText="Add previously built or imported criteria logic to branch node."
-      />
+      {useCriteriaSelected && (
+        <>
+          <DropDown
+            id="Criteria"
+            label="Criteria"
+            options={criteriaOptions}
+            onChange={selectCriteriaSource}
+            value={transition.condition?.cql || undefined}
```
blangley28, line 99: What happens here if `transition.condition.cql` is not in `criteriaOptions`?

src/components/Sidebar/BranchTransition.tsx
```diff
@@ -27,12 +80,38 @@ const BranchTransition: FC<BranchTransitionProps> = ({ pathway, transition, upda
...
+          <FormControl variant="outlined" fullWidth>
+            <TextField
+              label="Criteria display"
+              value={transition.condition?.description || ''}
+              variant="outlined"
+              onChange={setCriteriaDisplay}
+              error={
+                transition.condition?.description === undefined ||
+                transition.condition?.description === ''
+              }
```
blangley28, line 110: The empty string actually evaluates to false so you should be able to make this
```suggestion
              error={!transition.condition?.description}
```

src/components/Sidebar/BranchTransition.tsx
```diff
@@ -1,22 +1,75 @@
...
+          transition.condition?.description || '',
+          elm,
+          criteriaSource
+        )
+      );
+    },
+    [currentNodeKey, transition.id, updatePathway, pathway, transition.condition, criteria]
+  );
 
+  const setCriteriaDisplay = useCallback(
```
blangley28, line 62: Right now each criteria in the provider has only one statement. This would be a backlog item but what do you think about having this set a property in the criteria itself. Then if it already exists it would pre-populate the text field.

src/components/Sidebar/BranchTransition.tsx
```diff
@@ -1,22 +1,75 @@
...
 import { SidebarHeader, SidebarButton } from '.';
 import { Pathway, Transition } from 'pathways-model';
+import { useCriteriaContext } from 'components/CriteriaProvider';
 import useStyles from './styles';
 
 interface BranchTransitionProps {
   pathway: Pathway;
+  currentNodeKey: string;
   transition: Transition;
   updatePathway: (pathway: Pathway) => void;
```
blangley28, line 16: I think this is a good candidate to be brought into a provider. The function is defined in `BuilderRoute` then passed through `Builder` -> `Sidebar` -> `BranchNode` -> `BranchTransition`

src/components/Sidebar/BranchTransition.tsx
```diff
@@ -27,12 +80,38 @@ const BranchTransition: FC<BranchTransitionProps> = ({ pathway, transition, upda
...
+        <>
+          <DropDown
+            id="Criteria"
+            label="Criteria"
+            options={criteriaOptions}
+            onChange={selectCriteriaSource}
+            value={transition.condition?.cql || undefined}
+          />
+          <FormControl variant="outlined" fullWidth>
+            <TextField
```
blangley28, line 102: When I type into this textbox nothing happens. Unless is REALLY slow (like more than a minute) it's not working for me

src/components/Sidebar/BranchTransition.tsx
mdube7: The dropdown doesn't display a value

src/components/Sidebar/BranchTransition.tsx
noranda: Agreed, although I think this work is more in scope with [PATHWAYS-292](https://jira.mitre.org/browse/PATHWAYS-292).

src/components/Sidebar/BranchTransition.tsx
mdube7: That could be interesting, We would have to decide if something was already in the description but the criteria label was changing what happens.

src/components/Sidebar/BranchTransition.tsx
```diff
@@ -27,12 +80,38 @@ const BranchTransition: FC<BranchTransitionProps> = ({ pathway, transition, upda
...
+              label="Criteria display"
```
noranda, line 103:
```suggestion
              label="Criteria Display"
```

src/utils/builder.ts
mdube7: I ended up not needing this once I incorporated 261, going to remove it.

src/components/Sidebar/BranchTransition.tsx
noranda: Weird. It works for me on Mac/Chrome. This comment should probably be added to the optimization task ([PATHWAYS-292](https://jira.mitre.org/browse/PATHWAYS-292)).

mdube7:
Added Pathways-297 to disable the Use Criteria button when no criteria exists.

noranda:
Looks good! You do have a lint error though to remove the `FormControl` import since you removed it in `BranchTransition`.

src/components/Sidebar/BranchTransition.tsx
mdube7: Added this comment to Pathways-292

src/components/Sidebar/BranchTransition.tsx
blangley28: Yea that's a good point... I'll throw it to Jim & Noranda and see what the use case is. I assumed it would always be the same but I guess that may be too strict. There's nothing wrong with how you're doing it here but if the label is always the same then it would potentially save some duplicate typing

noranda:
Per @radamson's comment above on changing the node name, I recommend going with his first solution. To do so, add this to the `Input` component in `SidebarHeader`:
```js
onFocus={(event: FocusEvent<HTMLInputElement>): void => event.target.select()}
```

PR #13, opened 2020-06-08
Null ClientWidth on Resize

blangley28:
Resizing the window would cause an error because the nodeRef was null. I played around with it a lot and noticed whenever the screen resized the nodes were being set to null. I'm not exactly sure why but adding a check so the ref isn't updated when it is null seems to fix the issue.

To test create a new pathway and add at least one new node. Then resize the screen. No error is thrown and the nodes stay centered in the window.

### DIFF
--- a/src/components/Graph/Graph.tsx
+++ b/src/components/Graph/Graph.tsx
@@ -221,6 +221,7 @@ const GraphMemo: FC<GraphMemoProps> = memo(
                   name={nodeName}
                   ref={(node: HTMLDivElement): void => {
                     if (node) nodeRefs.current[nodeName] = node;
+                    else delete nodeRefs.current[nodeName];
                   }}
                   pathwayState={pathway.states[nodeName]}
                   isCurrentNode={false}
--- a/src/components/Graph/Graph.tsx
+++ b/src/components/Graph/Graph.tsx
@@ -43,14 +43,18 @@ const Graph: FC<GraphProps> = memo(({ pathway, interactive = true, expandCurrent
     if (nodeRefs?.current) {
       Object.keys(nodeRefs.current).forEach(key => {
         const nodeElement = nodeRefs.current[key];
-        const width = nodeElement.clientWidth;
-        // nodeElement can have multiple children so calculate the sum to get the node height
-        const height = Array.from(nodeElement.children).reduce(
-          (acc, child) => acc + child.clientHeight,
-          0
-        );
+        if (nodeElement) {
+          const width = nodeElement.clientWidth;
+          // nodeElement can have multiple children so calculate the sum to get the node height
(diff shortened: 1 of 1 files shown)


dehall:
So, this works, but I'm a little concerned that it's just covering up something else that's wrong. I think it's worth spending just a little more time trying to understand what's going on. 

Not sure if this is our exact issue but there are some interesting responses here: https://stackoverflow.com/questions/38401879/react-losing-ref-values

radamson:
It might be better to put this null check near where the ref is used (instead of where it is set).  That way if the referenced component unmounts for any other reason in the future we won't have this problem.

It looks like this is how the examples in the [React Docs](https://reactjs.org/docs/refs-and-the-dom.html#callback-refs) do it as well.  i.e. the `if (this.textInput)` check is in `focusTextInput` not `setTextInput`.

In this case we could wrap the code in the [`getGraphLayout` callback to make sure nodeElement](https://github.com/mcode/pathway-builder/blob/master/src/components/Graph/Graph.tsx#L45) isn't null before trying to use it.  There could certainly be more going on here though 🙂 .

src/components/Graph/Graph.tsx
```diff
@@ -220,7 +220,7 @@ const GraphMemo: FC<GraphMemoProps> = memo(
                   key={nodeName}
                   name={nodeName}
                   ref={(node: HTMLDivElement): void => {
-                    nodeRefs.current[nodeName] = node;
+                    if (node) nodeRefs.current[nodeName] = node;
```
noranda, line 223: @radamson's suggestion is valid, although I would suggest we go a step further.

The reason `node` is null is that the callback function passed to `ref` is called every time the ref changes. This happens on mount and unmount. On unmount the element is destroyed so we get `null`. If we only assign the node if present, we end up holding onto the `div` tag that is no longer on the page. This div will now be used in the `getGraphLayout` calculation, even though it’s no longer on the page. This also creates a memory leak, where the browser cannot release the memory used by that div because we are holding onto it.

Instead, if we change it to delete the key out of the `nodeRefs.current` we can continue using `Object.keys(nodeRefs.current)` without the need for any null checks, as every key should have a value.

```suggestion
                    if (node) {
                      nodeRefs.current[nodeName] = node;
                    } else {
                      delete nodeRefs.current[nodeName];
                    }
```

blangley28: That makes sense. Implemented

radamson:
This fix looks good to me.  Still think something else might be going on because I could only replicate it when adding a new node (I can resize fine with the existing pathways and their nodes).  But this fixes the issue and it would probably be better investigated with PATHWAYS-265 (i.e. in a larger refactor of Graph)

PR #14, opened 2020-06-10
add action node

KeeyanGhoreshi:
i think this gets the gist.  Add an action node, get an action node.


src/components/Sidebar/ActionNode.tsx
```diff
@@ -1,49 +1,212 @@
...
+            resourceType: actionType?.label,
+            title: ''
+          }
+        };
+      } else if (
+        currentState.action.length > 0 &&
+        currentState.action[0].resource.resourceType !== 'Careplan'
+      ) {
+        action = currentState.action[0];
+        action.resource.resourceType = actionType?.label;
```
blangley28, line 103: Did the CarePlan replace the need for multiple actions? If so then we should probably update the data model to make Action a single object instead of a list.
The reason I bring this up is because it looks like this block means we cannot add multiple actions - it just replaces the action type of the first one

src/components/Sidebar/ActionNode.tsx
```diff
@@ -1,49 +1,212 @@
...
   );
+  const currentNodeKey = currentNode?.key;
+
+  const changeCode = useCallback(
+    (event: ChangeEvent<{ value: string }>): void => {
+      if (!currentNodeKey) return;
+
+      const code = event?.target.value || '';
+      const currentState: GuidanceState = pathway.states[currentNodeKey] as GuidanceState;
+      const action: Action = currentState.action[0];
```
blangley28, line 59: We should probably put something in the code style guide to keep everything consistent but you don't need the types here
```suggestion
      const currentState = pathway.states[currentNodeKey] as GuidanceState;
      const action = currentState.action[0];
```

src/components/Sidebar/ActionNode.tsx
```diff
@@ -1,49 +1,212 @@
...
+      action.resource.code.coding[0].system = codeSystem;
+      updatePathway(setStateAction(pathway, currentNodeKey, [action]));
+    },
+    [currentNodeKey, pathway, updatePathway]
+  );
   // The node has a key and is not the start node
   const changeNodeTypeEnabled = currentNode.key !== undefined && currentNode.key !== 'Start';
 
+  const action = currentNode.action;
+  const resource = action && action.length > 0 && action[0].resource;
```
blangley28, line 145: The null check for action can be added to the second statement
```suggestion
  const resource = action?.length > 0 && action[0].resource;
```

src/components/Sidebar/ActionNode.tsx
```diff
@@ -1,49 +1,212 @@
...
+              />
+              {resource.code.coding[0].system && (
+                <Input
+                  id="code"
+                  label="Code"
+                  onChange={changeCode}
+                  value={resource.code.coding[0].code}
+                />
+              )}
+              {resource.code.coding[0].system && (
```
blangley28, line 184: If this was intended to be hidden until a system is selected then is there a reason it can't be included with the `Input` above it? Perhaps it should be hidden until a code is set though

src/components/Sidebar/ActionNode.tsx
```diff
@@ -1,49 +1,212 @@
...
+  updatePathway
+}) => {
   const styles = useStyles();
   const selectNodeType = useCallback(
     (event: ChangeEvent<{ value: string }>): void => {
       changeNodeType(event?.target.value || '');
     },
     [changeNodeType]
   );
+  const currentNodeKey = currentNode?.key;
```
blangley28, line 51: We should be consistent with using `currentNode` and its key:
1. For starters I don't think `currentNode` can be null so we shouldn't need to check for that
2. This also means all of the `if (!currentNodeKey) return;` statements can be removed
3. The `currentNode` is already passed as a prop but in multiple places you use the key to find the node from the pathway instead of just using the node passed in.
4. Some places use `currentNodeKey` but others use `currentNode.key`. I don't really mind which one we use but it should be consistent.

src/types/pathways-model.d.ts
```diff
@@ -37,19 +37,20 @@ declare module 'pathways-model' {
     cql: string;
     elm?: ElmLibrary;
     action: Action[];
+    title?: string;
+    description?: string;
```
blangley28, line 41: What are these used for?

src/utils/builder.ts
```diff
@@ -268,6 +268,32 @@ export function setStateCriteriaSource(
...
+    states: {
+      ...pathway.states,
+      [key]: {
+        ...pathway.states[key],
+        action
+      }
+    }
+  };
+}
+
```
blangley28, line 296: This method is identical to the one above

src/utils/builder.ts
```diff
@@ -268,6 +268,32 @@ export function setStateCriteriaSource(
   };
 }
 
+export function setStateAction(pathway: Pathway, key: string, action: Action[]): Pathway {
```
blangley28, line 271: When a guidance state is created it initialized the action to be an empty list. I suggest making this method `addStateAction` to create a new action. Then use the other set action methods to set the specific attributes of the action

src/components/Sidebar/ActionNode.tsx
dehall: I told Keeyan separately that since we only ever show one action at a time, for starters this PR should use the existing data model and only put one item in the list. If we want to change the data model, which I would be good with, let's do that separately.

KeeyanGhoreshi: I'm not sure what to do here since the designs only show one action being added but the data model allows for multiple.  I've talked to dylan and we decided to keep the data model the same even though it only allows 1 action for now.

src/components/Sidebar/ActionNode.tsx
KeeyanGhoreshi: currentNodeKey can be undefined, so we still have to check it.  Other stuff I agree.

src/utils/builder.ts
KeeyanGhoreshi: isn't this already accomplished inside the action node?  What benefit do we get from swapping stuff around like this?

blangley28:
When I select an action node is doesn't pre-populate the resource (action) type
Screen Shot 2020-06-15 at 12 15 08 PM
But once I manually select the correct type it populates the coding
Screen Shot 2020-06-15 at 12 16 17 PM

src/components/Sidebar/ActionNode.tsx
```diff
@@ -1,49 +1,212 @@
...
 
 const nodeTypeOptions = [
   { label: 'Action', value: 'action' },
   { label: 'Branch', value: 'branch' }
 ];
 
+const actionTypeOptions = [
+  { label: 'Medication Request', value: 'Medication Request' },
+  { label: 'Service Request', value: 'Service Request' },
+  { label: 'Careplan', value: 'Careplan' }
```
blangley28, line 20: The values here are being used in the `resource` but are not the correct fhirnames
```suggestion
  { label: 'Medication Request', value: 'MedicationRequest' },
  { label: 'Service Request', value: 'ServiceRequest' },
  { label: 'Careplan', value: 'CarePlan' }
```

src/components/Sidebar/ActionNode.tsx
```diff
@@ -1,49 +1,212 @@
...
+  const currentNodeKey = currentNode?.key;
+
+  const changeCode = useCallback(
+    (event: ChangeEvent<{ value: string }>): void => {
+      if (!currentNodeKey) return;
+
+      const code = event?.target.value || '';
+      const currentState: GuidanceState = pathway.states[currentNodeKey] as GuidanceState;
+      const action: Action = currentState.action[0];
+      action.resource.code.coding[0].code = code;
```
blangley28, line 60: The code isn't always `resource.code.coding[0]`. For `MedicationRequest` its `resource.medicationCodeableConcept`

noranda:
Designs for description field:
https://invis.io/6TX9T30J3M4#/418719440_Camino-Builder-Action-05

blangley28:
I don't think this is an issue with just PR but I wanted to bring it up anyway - selecting from autofill in the text box looks really ugly. 
Screen Shot 2020-06-17 at 1 42 30 PM

Also from me filling out this form a lot on my branch it would be nice if after filling in a code you could click the "Validate" button or enter. In my branch I updated so the description field is hidden until it is "validated"

dehall:
What about the Add Action/Branch buttons at the bottom of the sidebar? https://projects.invisionapp.com/share/6TX9T30J3M4#/screens/421770938
I'm willing to create a new item for that if it's a big addition, but if it's easy and we can work that in here that would be great.

blangley28:
> What about the Add Action/Branch buttons at the bottom of the sidebar? 
I don't follow - what about them? Have them hidden until all of the information is added?

dehall:
Sorry I should have a) tested more before commenting and b) been more clear. On a node that has an existing outgoing transition, nothing shows up in the sidebar to add a new option or edit the current transition: 
image

On a node that doesn't have existing transitions the buttons do show up as expected: 
image

I think ultimately what we want in that first case is something similar to the branch node transitions editor, but without the CQL criteria parts. We still want to allow multiple transitions for when there are multiple options to choose. Let's make that a separate task so we can include more designs on how transitions should work.

noranda:
Issues remaining:

1. When an action node is first selected (click on "add action node"), there should be no node icon. Only when a user selects a node type should the corresponding icon appear in the node header.

2. Is there a reason we took out ICD-9 and ICD-10 code systems? For reference, CDS Connect uses the following code systems:
```js
const codeSystemOptions = [
  { value: 'SNOMED', label: 'SNOMED', id: 'http://snomed.info/sct' },
  { value: 'ICD-9-CM', label: 'ICD-9-CM', id: 'http://hl7.org/fhir/sid/icd-9-cm' },
  { value: 'ICD-10-CM', label: 'ICD-10-CM', id: 'http://hl7.org/fhir/sid/icd-10-cm' },
  { value: 'NCI', label: 'NCI', id: 'http://ncimeta.nci.nih.gov' },
  { value: 'LOINC', label: 'LOINC', id: 'http://loinc.org' },
  { value: 'RXNORM', label: 'RXNORM', id: 'http://www.nlm.nih.gov/research/umls/rxnorm' },
  { value: 'Other', label: 'Other', id: null }
];
```

3. As soon as an action type is selected (ie. the node now includes metadata), the current node should expand in the graph to show this.

4. Remove the Input component as it is not used anywhere.

5. The Action Type in the sidebar no longer matches the Type in the node. Recommend we change the Type in the node to include the new user-friendly types. Ex. "Procedure (ServiceRequest)" or just "Procedure".

image

6. Regarding @blangley28's comment on the autofill styles, add this to the theme file on line 112 to fix:

```js
...
MuiInputBase: {
    ...,
    input: {
      ...,
      '&:-webkit-autofill': {
        transitionDelay: '9999s',
        transitionProperty: 'background-color, color'
      }
    }
  }
...
```

7. Remove the top margin, bold text, and fix font size on the display text to match the design.

Screen Shot 2020-06-19 at 12 52 04 PM

Design:
Screen Shot 2020-06-19 at 12 53 43 PM

KeeyanGhoreshi:
I think I addressed all of the points in the latest commit(s).  Though for (3) I wasn't sure how exactly to get the Action Node to expand when it first got metadata.  Since assigning the type occurs in the `ActionNode`, and handling of expansion occurs in the `Node`, there's not really a good way to communicate to `Node` that a certain action has been fired in `ActionNode`.  

Since the action node can only really update the graph by changing the pathway's state, I just used the pathways state to check if the state changed from having no metadata to having some metadata.  But this solution seemed a bit hacky to me.  Ideally the node would expand as a result of the function in `ActionNode` populating the metadata, not as a side effect of checking the state every time to see if the metadata has been updated. 

I'm open to any ideas on how this could be improved, if there are any.

blangley28:
Not sure if this is being addressed but I just rebased my branch off your latest changes I noticed the following
1. Clicking enter to validate doesn't work for regimens
2. Validating a regimen gives an error. 
>ActionNode.tsx:188 Uncaught TypeError: Cannot read property 'coding' of undefined
Screen Shot 2020-06-22 at 6 49 15 PM

src/components/ExpandedNode/ExpandedNode.tsx
```diff
@@ -28,10 +28,14 @@ type ExpandedNodeFieldProps = {
 
 const ExpandedNodeField: FC<ExpandedNodeFieldProps> = ({ title, description }) => {
   return (
-    <tr>
-      <td className={styles.descTitle}>{title}</td>
-      <td className={styles.desc}>{description}</td>
-    </tr>
+    <>
+      {description ? (
```
noranda, line 32: This can be better written as `{description && ( ... )}` if the else is null.

noranda:
Issues remaining:

1. An error is thrown when typing a Regimen title and then deleting it.

Screen Shot 2020-06-23 at 10 30 58 AM

2. I noticed for regimens/care plans, we are now inputing a title as opposed to a code and system, yet we are still displaying the option to validate. Are we able to validate this? If so, no changes needed. If not, we should remove the validate button.

Screen Shot 2020-06-23 at 10 33 19 AM

noranda:
I found one last issue. To reproduce:

1. Create a new pathway
2. Add an action node
3. Expand the action node in the graph
4. Add a type to the action node

This will collapse the node which is not the behavior we want. The node should remain open, not toggle, when metadata exists. Everything else looks good to me.

PR #15, opened 2020-06-11
Current nodes are now red and the choice nodes and choice arrows stem…

Jack-Fraser16:
…ming from current branch nodes are similarly red. Known issue: different nodes with the same label i.e. Surgery are both red when one is clicked. This is due to the fact that the currentNode is  matched via label against all other nodes. I have some ideas on how to fix this, but I wanted to get feedback on code organization first. Passing down the currentNode object from builder.tsx feels a little clunky. Is there a better way to determine which node is current? Maybe something that can be accessed from within the graph component without props? Just an idea, let me know your thoughts!


src/components/Arrow/Arrow.module.scss
```diff
@@ -20,3 +20,7 @@
 .arrow {
   @include arrow($blue);
 }
+
+.branchArrow {
+  @include arrow($red);
+}
```
blangley28, line 26: The naming here is a little confusing. Its not that all branch arrows are red but rather selected/current/highlighted branch arrows are red. 
EDIT: it looks like this naming convention carries down into JS variables as well which should be updated

src/components/Graph/Graph.tsx
```diff
@@ -223,11 +241,12 @@ const GraphMemo: FC<GraphMemoProps> = memo(
                     nodeRefs.current[nodeName] = node;
                   }}
                   pathwayState={pathway.states[nodeName]}
-                  isCurrentNode={false}
+                  isCurrentNode={pathway.states[nodeName].label === currentNode?.label}
```
blangley28, line 244: As you mentioned in the comments the label is not unique but the key is. Since we are dealing with the entire nodes here and not just passing around the label I think the only change needed is:
```suggestion
                  isCurrentNode={pathway.states[nodeName].key === currentNode?.key}
```

src/components/Graph/Graph.tsx
```diff
@@ -223,11 +241,12 @@ const GraphMemo: FC<GraphMemoProps> = memo(
...
                   xCoordinate={nodeCoordinates[nodeName].x + parentWidth / 2}
                   yCoordinate={nodeCoordinates[nodeName].y}
                   expanded={Boolean(expanded[nodeName])}
                   onClick={onClickHandler}
+                  isChoiceOfCurrent={isBranchState(currentNode) && childOfCurrent}
```
blangley28, line 249: Another naming comment - to stay consistent with current naming conventions this would be if the node is a transition of the current node so I would rename to `isTransitionOfCurrentBranch`

src/components/Graph/Graph.tsx
```diff
@@ -215,6 +222,17 @@ const GraphMemo: FC<GraphMemoProps> = memo(
       >
         {nodeCoordinates !== undefined
           ? Object.keys(nodeCoordinates).map(nodeName => {
+              let childOfCurrent = false;
```
blangley28, line 225: Another naming comment - to stay consistent with current naming conventions this would be if the node is a transition of the current node so I would rename to `transitionOfCurrent`

Jack-Fraser16:
Everything should be good to go !

src/components/Node/Node.tsx
```diff
@@ -42,11 +42,15 @@ const Node: FC<NodeProps & { ref: Ref<HTMLDivElement> }> = memo(
         left: xCoordinate
       };
 
+      const isCurrentNode = pathwayState.key === currentNode?.key;
```
blangley28, line 45: On line 26 currentNode is defined as 
`currentNode: state`
So it should never be null. Therefore you won't need the null check here
```suggestion
      const isCurrentNode = pathwayState.key === currentNode.key;
```

src/components/Node/Node.tsx
```diff
@@ -19,17 +19,17 @@ import {
...
   expanded?: boolean;
   onClick?: (nodeName: string) => void;
+  currentNode: State;
 }
 
 const Node: FC<NodeProps & { ref: Ref<HTMLDivElement> }> = memo(
   forwardRef<HTMLDivElement, NodeProps>(
     (
-      { name, pathwayState, isCurrentNode, xCoordinate, yCoordinate, expanded = false, onClick },
+      { name, pathwayState, xCoordinate, yCoordinate, expanded = false, onClick, currentNode },
```
blangley28, line 32: I know this wasn't you but can we rename `name` here in these props. It is really the `nodeKey`

noranda:
This is looking pretty good! I was able to find one last bug that should be addressed. To reproduce:

1. Create a new pathway
2. Add branch node
3. Add transition

Notice here that the new transition node is not red. However, when you add a second transition node, they both are now red as expected.

Screen Shot 2020-06-22 at 4 20 04 PM

blangley28:
Almost ready to approve - everything is in order for me except when I was doing one last test on the app I noticed the following:

Behavior is as expected on existing pathway but when creating a new pathway the _first_ transition is not highlighted. Once a second transition is added then they both are.
Screen Shot 2020-06-23 at 11 28 44 AM
Screen Shot 2020-06-23 at 11 28 58 AM

PR #16, opened 2020-06-12
Fix arrow rendering

danlee1025:
Modified the method we used to calculate `maxWidth`. Previously we depended on the edge label farthest to the right to determine the width for the arrow svgs so the arrows were not appearing when there were no edge labels. I changed it to use the node farthest to right instead.

image

### DIFF
--- a/src/components/Graph/Graph.tsx
+++ b/src/components/Graph/Graph.tsx
@@ -94,8 +94,8 @@ const Graph: FC<GraphProps> = memo(({ pathway, interactive = true, expandCurrent
   const maxWidth = useMemo(() => {
     return nodeCoordinates !== undefined
       ? Object.values(nodeCoordinates)
-          // Add 250 to account for expanded node width
-          .map(x => x.x + parentWidth / 2 + 250)
+          // Add width of the node to account for x coordinate starting at top left corner
+          .map(x => x.x + parentWidth / 2 + (x.width ?? 0))
           .reduce((a, b) => Math.max(a, b))
       : 0;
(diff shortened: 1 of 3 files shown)


danlee1025:
@mdube7 I'm assuming it has to do with PATHWAYS-272 so I did not look at it.

src/components/Graph/Graph.tsx
```diff
@@ -90,6 +90,16 @@ const Graph: FC<GraphProps> = memo(({ pathway, interactive = true, expandCurrent
     });
   }
 
+  // Find node that is farthest to the right
+  const maxWidth = useMemo(() => {
+    return nodeCoordinates !== undefined
+      ? Object.values(nodeCoordinates)
+          // Add 250 to account for expanded node width
+          .map(x => x.x + parentWidth / 2 + 250)
```
dehall, line 98: Is 250 already defined as or based on a constant somewhere? Or is this just based on experimentation? The comment is good but if we can link it to something more concrete that would be good.

danlee1025: Mostly based on experimentation. The min width for an expanded node is 400 so 250 was chosen to cover arrows drawn slightly past the halfway point. I'll try to find something more concrete to link it to.

danlee1025: Ok I changed this to use the width of the node instead of the random 250 number.

dehall:
Travis 😞 
It passed though: https://travis-ci.org/github/mcode/pathway-builder/builds/699434085

PR #17, opened 2020-06-19
Make the start node read only and select icon based on the key

mdube7:
This addresses the issue of the start node having the wrong icon.

### DIFF
--- a/src/components/Graph/Graph.tsx
+++ b/src/components/Graph/Graph.tsx
@@ -111,6 +111,7 @@ const Graph: FC<GraphProps> = memo(({ pathway, interactive = true, expandCurrent
   );
 
   const toggleExpanded = useCallback((key: string) => {
+    if (key === 'Start') return;
     setExpanded(prevState => ({
       ...prevState,
       [key]:
--- a/src/components/Graph/Graph.tsx
+++ b/src/components/Graph/Graph.tsx
@@ -111,13 +111,19 @@ const Graph: FC<GraphProps> = memo(({ pathway, interactive = true, expandCurrent
   );
 
   const toggleExpanded = useCallback((key: string) => {
-    if (key === 'Start') return;
-    setExpanded(prevState => ({
-      ...prevState,
-      [key]:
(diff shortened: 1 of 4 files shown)


src/components/Sidebar/SidebarHeader.tsx
```diff
@@ -85,13 +85,16 @@ const SidebarHeader: FC<SidebarHeaderProps> = ({
                 onKeyPress={handleKeyPress}
                 defaultValue={currentNodeLabel}
                 autoFocus
+                readOnly={currentNodeKey === 'Start'}
```
noranda, line 88: This makes the input read only, but we really don't want the input to show up at all if it can't be edited. To fix, remove this line, and add `&& currentNodeKey !== 'Start'` to the check on line 78.

We also want to get rid of the pointer cursor since it is no longer interactive. To do so, change the className on the header label div on line 93 to:

```js
className={clsx(
  styles.headerLabel,
  styles.headerLabelText,
  currentNodeKey === 'Start' && styles.headerLabelStart
)}
```

Then add the following style to the sidebar stylesheet: 

```js
headerLabelStart: {
  cursor: 'default'
}
```

mdube7: Prevent start from expanding and made these changes.

src/components/Graph/Graph.tsx
```diff
@@ -111,6 +111,7 @@ const Graph: FC<GraphProps> = memo(({ pathway, interactive = true, expandCurrent
   );
 
   const toggleExpanded = useCallback((key: string) => {
+    if (key === 'Start') return;
```
dehall, line 114: I think this line is causing some weirdness with the expand/unexpand logic when switching back & forth between nodes. It's definitely a super corner case to switch back & forth between Start and anything else, but if you have Node B open, click Start, then click back to Node B, it closes B instead of keeping it open. (It does work correctly for any other pair of nodes that don't include Start). If there's an easy fix for that let's make it here, otherwise I'd suggest reverting this line and making a new task to keep the start node from expanding.

PR #18, opened 2020-06-22
Generate CQL for Action States

blangley28:
Generates the CQL automatically for action states based on the action resources.

Few notes:
1. This is based off Keeyan's branch so I set up the PR to have that as the base. Right now some things might be broken as he continues to update that branch.
2. The ELM is all valid - you will see annotation warnings about the retrieve. I passed these buy Chris as the ELM does what is expected. He said the ELM is correct but the translator is adding that warning by mistake
3. I added a sample pathway which represents a pathway right after you create a new one. This way for testing we no longer need to create a new pathway everytime we can use that one which has a static uri
4. I broke out the ELM data models into their own file as they were getting more and more complex


src/utils/builder.ts
```diff
@@ -503,6 +604,53 @@ export function makeStateBranch(pathway: Pathway, stateKey: string): Pathway {
   };
 }
 
+export function createCQL(action: Action, stateKey: string): string {
```
dehall, line 607: I'm a little confused by the content of this pull request... I know it's based on Keeyan's branch but the delta seems to be including a lot of UI items from that branch for some reason. For now though I'm assuming this function is the majority of what's relevant to look at here.

blangley28: I think he may have pushed again so I need to rebase again

src/types/elm-model.d.ts
```diff
@@ -0,0 +1,88 @@
+declare module 'elm-model' {
```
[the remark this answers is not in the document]
blangley28: This is almost certainly NOT the entire ELM data model. I couldn't find a model anywhere and Moesel also said he wasn't aware of one. I knew this wouldn't be everything which is why at the end I just added `[x: string]: object;` to catch anything else inside the object.

src/components/Sidebar/ActionNode.tsx
```diff
@@ -52,6 +59,17 @@ const ActionNode: FC<ActionNodeProps> = ({
     [changeNodeType]
   );
 
+  const addActionCQL = useCallback(
+    (action: Action, currentNodeKey: string): void => {
+      const cql = createCQL(action, currentNodeKey);
+      convertBasicCQL(cql).then(elm => {
+        // eslint-disable-next-line
```
danlee1025, line 66: How come this is needed? I don't think I see a lint error when I remove it.

blangley28: Oh yea you're right that can be removed now. Before this was its own method it was using `currentNode.key` not `currentNodeKey`. A check was already done to ensure the key was defined but the anonymous function wasn't smart enough to know that. As such I had to use `currentNode.key!` which gave a lint error

danlee1025:
This looks really good to me, are you finished with the fixes you mentioned at standup?

blangley28:
> are you finished with the fixes you mentioned at standup?

Yup! just pushed two more styling changes Noranda suggested

dehall:
I'm getting a couple errors when I try things with Regimens. Both cases use the Early Stage HER2+ pathway and click on the No Endocrine Therapy node to start.

Case 1: Add a Regimen here, give it a name and description, then export the pathway. This error appears:
image


Case 2: Add a new Action node, and just type really fast or delete the name field. This one I can't reproduce as consistently but eventually this error appears:
image

danlee1025:
For Case 1, I'm pretty sure we just need to add a null check, I did something similar when working on export pathway.

blangley28:
> For Case 1, I'm pretty sure we just need to add a null check, I did something similar when working on export pathway.

yea that sounds like it would work - however I think this is a slightly bigger issue of my elm model requiring all ofthese elements when really they should be optional. I'll spend some time updating

blangley28:
Fixed the two bugs from Dylan and also found this one which occurs when the code or title includes `'`. Because the CQL wraps these variables in `'` it must be escaped. In this PR I used the JS `escape` method. Adding a backlog item to research if this has any impact to using this CQL in Camino (does Camino also need to escape the titles and codes so it is a match)

Screen Shot 2020-06-26 at 12 13 55 PM

PR #20, opened 2020-06-23
Default criteria uploaded.

Jack-Fraser16:


Jack-Fraser16:
Screen Shot 2020-06-23 at 4 33 21 PM

Jack-Fraser16:
I basically copied and reworked some of the code in the CriteriaProvider and pulled out the portion dealing with JSON to criteria parsing as an independent function. It was the easiest way to deal with local JSON files.

src/components/CriteriaService/CriteriaService.tsx
```diff
@@ -0,0 +1,41 @@
+import { useState, useEffect } from 'react';
```
blangley28, line 1: I don't have a ton of exposure to the PathwayService but is there anyway we could generalize that and share instead of copying everything here and just changing types?

src/components/PathwayProvider.tsx
```diff
@@ -48,10 +48,16 @@ export const PathwayProvider: FC<PathwayProviderProps> = memo(({ children }) =>
     ]);
   }, []);
 
+  console.log('Pathway Payload');
+  console.log(servicePayload);
+
```
blangley28, line 53: Remove these debug statements
```suggestion
```

src/components/PathwayProvider.tsx
```diff
@@ -48,10 +48,16 @@ export const PathwayProvider: FC<PathwayProviderProps> = memo(({ children }) =>
...
   useEffect(() => {
     if (servicePayload) setPathways(servicePayload);
   }, [servicePayload]);
 
+  console.log('Pathway Payload2');
+  console.log(servicePayload);
+
```
blangley28, line 60: And these too
```suggestion
```

src/components/CriteriaService/CriteriaService.tsx
Jack-Fraser16: I'm thinking about generic typing the PathwayService module as a solution, but am open to other ideas!

src/components/PathwayProvider.tsx
```diff
@@ -29,7 +29,8 @@ interface PathwayProviderProps {
 
 export const PathwayProvider: FC<PathwayProviderProps> = memo(({ children }) => {
   const [pathways, setPathways] = useState<Pathway[]>([]);
-  const service = useGetPathwaysService(config.get('demoPathwaysService'));
+  const pathwayObject = {} as Pathway;
```
dehall, line 32: The approach of creating a dummy object to pass into the generic service feels a little strange, but I'm ok with it if necessary. I do think because TypeScript is bolted on instead of being a core part of the language, we can't do things as cleanly as we might like. That said, if it's possible to either 
A) Add generic typing to the useGetService function signature, so it would look like `useGetService<Pathway>(...)`
or
B) pass in the type `Pathway` directly instead of an instance of it
Either of those would be ideal IMO.

Jack-Fraser16: Yeah, totally agree. I think I can figure out option (A).

PR #22, opened 2020-06-30
Allows users to edit the metadata for any existing pathway. Additiona…

Jack-Fraser16:
…lly, pathway name is now a button to access the builder view.

### DIFF
--- a/src/components/PathwaysList/PathwayModal.tsx
+++ b/src/components/PathwaysList/PathwayModal.tsx
@@ -90,7 +90,6 @@ const PathwayModal: FC<PathwayModalProps> = ({ open, onClose, editPathway }) => 
             required={createNewPathwayMeta}
             inputRef={pathwayNameRef}
             defaultValue={createNewPathwayMeta ? undefined : name}
-            multiline
           />
 
           <TextField
--- a/src/components/PathwaysList/PathwayModal.tsx
+++ b/src/components/PathwaysList/PathwayModal.tsx
@@ -62,7 +62,7 @@ const PathwayModal: FC<PathwayModalProps> = ({ open, onClose, editPathway }) => 
         onClose();
       }
     },
-    [ updatePathway, editPathway, onClose]
+    [updatePathway, editPathway, onClose]
   );
 
   let name, description;
--- a/src/components/PathwaysList/PathwayModal.tsx
+++ b/src/components/PathwaysList/PathwayModal.tsx
@@ -16,18 +16,21 @@ import shortid from 'shortid';
 import { usePathwayContext } from 'components/PathwayProvider';
 import useStyles from './styles';
 import { createNewPathway } from 'utils/builder';
+import { Pathway } from 'pathways-model';
 
-interface NewPathwayModalProps {
+interface PathwayModalProps {
   open: boolean;
   onClose: () => void;
+  editPath?: Pathway;
 }
 
-const NewPathwayModal: FC<NewPathwayModalProps> = ({ open, onClose }) => {
+const PathwayModal: FC<PathwayModalProps> = ({ open, onClose, editPath }) => {
+  const createNewModal = !editPath;
   const styles = useStyles();
   const history = useHistory();
   const pathwayNameRef = useRef<HTMLInputElement>(null);
   const pathwayDescRef = useRef<HTMLInputElement>(null);
-  const { addPathway } = usePathwayContext();
+  const { addPathway, updatePathwayAtIndex, pathways } = usePathwayContext();
 
   const closeModal = useCallback(
     (pathwayId: string): void => {
@@ -49,6 +52,30 @@ const NewPathwayModal: FC<NewPathwayModalProps> = ({ open, onClose }) => {
     [addPathway, closeModal]
   );
 
+  const handleUpdatePathway = useCallback(
+    (event: FormEvent<HTMLFormElement>): void => {
+      event.preventDefault();
+      if (editPath) {
+        if (pathwayNameRef.current?.value) editPath.name = pathwayNameRef.current?.value;
+        if (pathwayDescRef.current?.value !== '')
+          editPath.description = pathwayDescRef.current?.value;
+        const pathwayId = editPath.id;
+        const pathwayIndex = pathways.findIndex(pathway => pathway.id === pathwayId);
+        updatePathwayAtIndex(editPath, pathwayIndex);
+        onClose();
+      }
+    },
+    [pathways, updatePathwayAtIndex, editPath, onClose]
+  );
+
+  let label = 'Pathway Name';
+  let description = 'Pathway Description';
+  if (editPath) {
+    label = editPath.name;
+    const metaDescription = editPath.description;
+    if (metaDescription && metaDescription !== '') description = metaDescription;
+  }
+
   return (
     <Dialog open={open} onClose={onClose} aria-labelledby="create-pathway" fullWidth maxWidth="md">
       <DialogTitle disableTypography>
@@ -57,23 +84,18 @@ const NewPathwayModal: FC<NewPathwayModalProps> = ({ open, onClose }) => {
         </IconButton>
       </DialogTitle>
 
-      <form onSubmit={handleCreateNewPathway}>
+      <form onSubmit={createNewModal ? handleCreateNewPathway : handleUpdatePathway}>
         <DialogContent>
           <TextField
             variant="outlined"
             autoFocus
-            label="Pathway Name"
+            label={label}
             fullWidth
-            required
+            required={createNewModal}
             inputRef={pathwayNameRef}
           />
 
-          <TextField
-            variant="outlined"
-            label="Pathway Description"
-            fullWidth
-            inputRef={pathwayDescRef}
-          />
+          <TextField variant="outlined" label={description} fullWidth inputRef={pathwayDescRef} />
         </DialogContent>
 
         <DialogActions>
@@ -83,7 +105,7 @@ const NewPathwayModal: FC<NewPathwayModalProps> = ({ open, onClose }) => {
             startIcon={<FontAwesomeIcon icon={faPlus} />}
             type="submit"
           >
-            Create
+            {createNewModal ? 'Create' : 'Update Info'}
           </Button>
         </DialogActions>
       </form>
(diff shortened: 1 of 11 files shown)


Jack-Fraser16:
There are still some design discrepancies. Primarily, making the name of each pathway a button means that the lettering is all upper case instead of just capitalized. Additionally, the text seems to be a bit smaller than the inVision design shows. I tried some quick, simple fixes to address these issues but they didn't work and I wanted to make sure the functionality is nailed down before diving into the CSS stuff. Let me know your thoughts!
Screen Shot 2020-06-30 at 3 39 30 PM
Screen Shot 2020-06-30 at 3 39 26 PM

dehall:
I'm not too worried about the size of the text, but if it can be made to not be all-caps that would be good.

In term of functionality, it works and correctly updates the pathway, but the UX is a little unconventional.
When I open the edit box, I get something that looks like this, which is good:
image

But then when I edit one of the text boxes, the old value shifts upward to be the label, and the box becomes empty. 
image

So if I want to just change 1 letter or a small piece of the text, which I assume (without evidence) will be the most common case, I have to re-type the whole thing.
Is it possible to keep the labels as "Pathway Name" and "Description" as before, and pre-populate the boxes with the old value so they are easier to make quick adjustments to?

dehall:
Also, not a showstopper but there seems to be a lot of empty space at the bottom of the modal, which you can scroll down into: 
image

src/components/PathwaysList/PathwaysTable.tsx
```diff
@@ -14,69 +14,92 @@ import {
...
-          {pathways.map(pathway => (
-            <TableRow key={pathway.id}>
-              <TableCell component="th" scope="row">
-                {pathway.name}
-              </TableCell>
+          <TableBody>
+            {pathways.map(pathway => (
+              <TableRow key={pathway.id}>
+                <TableCell component="th" scope="row">
+                  <Button
```
dehall, line 54: To address the font size and capitalization concerns -- maybe instead of a `Button` with `component={Link}` should this just be a pure `Link` component?

blangley28: This is what I was thinking too. It shouldn't need to be a button

src/components/PathwaysList/PathwayModal.tsx
```diff
@@ -49,6 +52,30 @@ const NewPathwayModal: FC<NewPathwayModalProps> = ({ open, onClose }) => {
     [addPathway, closeModal]
   );
 
+  const handleUpdatePathway = useCallback(
+    (event: FormEvent<HTMLFormElement>): void => {
+      event.preventDefault();
+      if (editPath) {
+        if (pathwayNameRef.current?.value) editPath.name = pathwayNameRef.current?.value;
+        if (pathwayDescRef.current?.value !== '')
```
blangley28, line 60: Is there a reason the description is checked to be non-empty but the name isn't? I feel like if anything it should be the other way around but in reality both should probably be non-empty

src/components/PathwaysList/PathwayModal.tsx
```diff
@@ -49,6 +52,30 @@ const NewPathwayModal: FC<NewPathwayModalProps> = ({ open, onClose }) => {
...
+          editPath.description = pathwayDescRef.current?.value;
+        const pathwayId = editPath.id;
+        const pathwayIndex = pathways.findIndex(pathway => pathway.id === pathwayId);
+        updatePathwayAtIndex(editPath, pathwayIndex);
```
blangley28, line 64: I know this isn't something new in this PR but I'm thinking perhaps we should update this function to take in an ID instead of an index. We never have the index naturally so we're constantly copying and pasting this `findIndex` line. It would be cleaner to have the update function handle that for us.

src/components/PathwaysList/PathwayModal.tsx
```diff
@@ -49,6 +52,30 @@ const NewPathwayModal: FC<NewPathwayModalProps> = ({ open, onClose }) => {
...
+      }
+    },
+    [pathways, updatePathwayAtIndex, editPath, onClose]
+  );
+
+  let label = 'Pathway Name';
+  let description = 'Pathway Description';
+  if (editPath) {
+    label = editPath.name;
+    const metaDescription = editPath.description;
```
blangley28, line 75: As Dylan mentioned we don't want to update the labels here but rather set the values of the refs

src/components/PathwaysList/PathwayModal.tsx
```diff
@@ -49,6 +52,30 @@ const NewPathwayModal: FC<NewPathwayModalProps> = ({ open, onClose }) => {
     [addPathway, closeModal]
   );
 
+  const handleUpdatePathway = useCallback(
+    (event: FormEvent<HTMLFormElement>): void => {
+      event.preventDefault();
+      if (editPath) {
```
blangley28, line 58: Is this here just because of typescript warnings? It shouldn't be needed since this function is called from `createNewModal ? handleCreateNewPathway : handleUpdatePathway` and `createNewModal = !editPath`

src/components/PathwaysList/PathwayModal.tsx
```diff
@@ -16,18 +16,21 @@ import shortid from 'shortid';
 import { usePathwayContext } from 'components/PathwayProvider';
 import useStyles from './styles';
 import { createNewPathway } from 'utils/builder';
+import { Pathway } from 'pathways-model';
 
-interface NewPathwayModalProps {
+interface PathwayModalProps {
   open: boolean;
   onClose: () => void;
+  editPath?: Pathway;
```
blangley28, line 24: Suggest this be renamed to something which includes the entire name "pathway". The identifier "path" doesn't have meaning in the builder but in Camino "path" refers to the actual nodes the patient was on whereas "pathway" is the entire datastructure

src/components/PathwaysList/PathwayModal.tsx
```diff
@@ -16,18 +16,21 @@ import shortid from 'shortid';
...
 }
 
-const NewPathwayModal: FC<NewPathwayModalProps> = ({ open, onClose }) => {
+const PathwayModal: FC<PathwayModalProps> = ({ open, onClose, editPath }) => {
+  const createNewModal = !editPath;
```
blangley28, line 28: Suggest this be renamed to `createNewPathway` since the boolean is whether or not we are creating a new pathway not a new modal

src/components/PathwaysList/PathwaysTable.tsx
```diff
@@ -14,69 +14,92 @@ import {
...
 const PathwaysTable: FC = () => {
   const styles = useStyles();
   const { pathways, deletePathway } = usePathwayContext();
+  const [open, setOpen] = useState(false);
+  const [editablePathway, setEditablePathway] = useState<Pathway>();
+
+  function openEditPathwayModal(pathway: Pathway): void {
+    setOpen(true);
+    setEditablePathway(pathway);
+  }
```
blangley28, line 31: This should be a callback as well
```suggestion
  const openEditPathwayModal = useCallback((pathway: Pathway): void => {
    setOpen(true);
    setEditablePathway(pathway);
  }, []);
```

Jack-Fraser16:
> Also, not a showstopper but there seems to be a lot of empty space at the bottom of the modal, which you can scroll down into:
> image

Super weird, but this empty space and scrolling option doesn't occur when I'm running it.

blangley28:
> Super weird, but this empty space and scrolling option doesn't occur when I'm running it.
It's only on the update for me but I see the same as Dylan with the scrolling

dehall:
I see it on all three browsers, Firefox, Chrome, and Safari. Looks like it's a different amount of space for each pathway, like it's based on the length of something from the pathway content or something. But looking at the inspector, I can't actually tell why that extra space is there. There's nothing else in the containing div besides the 2 fields

src/components/PathwaysList/PathwayModal.tsx
Jack-Fraser16: Technically, editPath.name has type string and pathwayNameRef.current?.value has type string or undefined, so the conditional is just a way to silence the type checking error from potentially assigning. From a user perspective, I think that the name of the pathway should always be some string, so if the user tries to Update with nothing in the Pathway Name text box, then the previous Pathway Name is maintained to protect against never having a Pathway Name. I can't think of a scenario where the user would actually want their pathway to be completely nameless.

Jack-Fraser16: With the understanding that, because creating a pathway requires a name, there will always be a previous name.

src/components/PathwaysList/PathwayModal.tsx
Jack-Fraser16: Okay! I can either go through and fix it now or maybe we should create a new item for it?

blangley28: I think we can create a new task for it - just wanted to see if other people agreed with me first

src/components/PathwaysList/PathwayModal.tsx
Jack-Fraser16: Agree that it shouldn't be needed, but it doesn't compile without it. I couldn't figure out why the object was being flagged as possibly undefined. It might be because editPath is in the dependency list, so maybe it's type checking because of potential changes elsewhere to editPath?

src/components/PathwaysList/PathwayModal.tsx
Jack-Fraser16: Used createNewPathwayMeta instead because createNewPathway is a function call

noranda:
I'll let the other reviewers handle looking at the code, but from a UI perspective, this looks pretty good! I only have 2 minor changes to request:

1. Lets change the button text to just "Save" with no icon.

Screen Shot 2020-07-02 at 3 12 34 PM

2. Let's change the pathway description field to be `multiline` so that we don't have the description run off the field.

Screen Shot 2020-07-02 at 3 13 05 PM

Otherwise, looks good! Nice job!

src/components/PathwaysList/PathwayModal.tsx
Jack-Fraser16: Whoops already dealt with it :(

PR #23, opened 2020-07-01
Standardize nomenclature

KeeyanGhoreshi:
This is a refactoring PR, focused mainly on resolving conflicting variable/type names in the code.  This is the sister PR to the one on the Pathways repo.  The changes are basically the same, save for a few differences in the files unique to the builder.

The changes have been broken up into different commits.  Though the code compiles/runs and the tests pass, there might be some errant name change that broke something.

Please let me know if you disagree with any of the name changes.


dehall:
Ack, sorry I was confused when I responded to your question earlier. In the case of "criteria" -- actually anything that linked to the CriteriaProvider/Build Criteria/Add Criteria/etc, all of that is correctly named "criteria". It's only the "criteria" already within the pathway that should be renamed to "preconditions".
State -> Node is good, Guidance -> Action is good, but we do want to keep some of the things called "Criteria".

blangley28:
Couldn't find anywhere you missed with the three changes (State -> Node, Guidance -> Action, and Name -> Key). I know I brought up the discrepancy between transition and branch nodes. I did a quick look but doesn't look like there are any issues here - did you already check?
It would be great if we can get these standard names documented in the wiki pages

PR #24, opened 2020-07-02
Update use criteria

mdube7:
This uses the new design for use criteria. There is a button to cancel if no data has been entered or to delete if data has been entered.

### DIFF
--- a/src/components/Sidebar/BranchTransition.tsx
+++ b/src/components/Sidebar/BranchTransition.tsx
@@ -2,13 +2,19 @@ import React, { FC, memo, useState, useCallback, ChangeEvent } from 'react';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faPlus, faTools } from '@fortawesome/free-solid-svg-icons';
 import DropDown from 'components/elements/DropDown';
-import { TextField } from '@material-ui/core';
-import { setTransitionCondition, setTransitionConditionDescription } from 'utils/builder';
+import { TextField, Button } from '@material-ui/core';
+import {
+  setTransitionCondition,
+  setTransitionConditionDescription,
+  removeTransitionCondition
+} from 'utils/builder';
 import { SidebarHeader, SidebarButton } from '.';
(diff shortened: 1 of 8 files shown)


src/components/Sidebar/BranchTransition.tsx
```diff
@@ -28,10 +33,34 @@ const BranchTransition: FC<BranchTransitionProps> = ({
...
+      setUseCriteriaSelected(!useCriteriaSelected);
+    }
+  }, [
+    useCriteriaSelected,
+    currentNodeKey,
+    pathway,
+    transition.condition,
+    transition.id,
+    updatePathway
+  ]);
```
noranda, line 63: Since `currentNodeKey` will never be null, you can remove the check and simplify this function a bit.
```js
 const handleUseCriteria = useCallback((): void => {
    if (hasCriteria && transition.id) {
      // delete the transition criteria
      updatePathway(removeTransitionCondition(pathway, currentNodeKey, transition.id));
      setUseCriteriaSelected(false);
    } else {
      setUseCriteriaSelected(!useCriteriaSelected);
    }
  }, [useCriteriaSelected, pathway, currentNodeKey, transition.id, updatePathway, hasCriteria]);
```

mdube7:
1. Created PATHWAYS-316 for the label display issue
2. Updated the nodes to display no icon if no type is defined.

Jack-Fraser16:
> 1. Created PATHWAYS-316 for the label display issue
> 2. Updated the nodes to display no icon if no type is defined.

Jack-Fraser16:
Seems great! One thing I noticed was that the sizes of the arrows coming off the transition node change when the user clicks on the criteria tab of the criteria selector. 

Screen Shot 2020-07-07 at 11 21 03 AM

Screen Shot 2020-07-07 at 11 21 09 AM

It's nothing super important, but the small noticeable change is a tad distracting. I'm assuming this should probably be it's own item?

mdube7:
Yea the arrow seems to have 2 sizes depending on whether there is a label or not. If you type anything into the criteria display you can see the arrow expand as well.

PR #25, opened 2020-07-02
Removed the missing data popup from expanded branch nodes. Additional…

Jack-Fraser16:
…ly, removed the MissingDataPopup component from the src code folder as it seems that it wasn't being used anywhere else.


Jack-Fraser16:
The only aspect I wasn't sure of is if expanded nodes should still have that little whitespace beneath the node name, even when they're empty. Is it part of the design?

Screen Shot 2020-07-02 at 3 57 27 PM

PR #27, opened 2020-07-13
increase style priority

KeeyanGhoreshi:
Basically there's a discrepancy between the dev and production build, where the custom CSS we wrote would be injected after the CSS for material UI.  This meant that for some components, our styles got overridden.

I fixed this by using the "index" property of the options for `makeStyles` to increase the priority of our css over that of material-ui, so it renders correctly on production builds.  There are other ways to solve this problem that involve forcing our CSS to inject before the material UI CSS, but they are a bit more complicated and require some more changes.  I went ahead and added an index to every style file, so that this problem does not happen in the future.  

You can see that this works by visiting my fork of the project [here](https://github.com/KeeyanGhoreshi/pathway-builder/tree/gh-pages).  The buttons looked like this:

Screen Shot 2020-06-24 at 10 05 23 AM

But have been fixed on my forked version by changing the index.

### DIFF
--- a/src/components/Arrow/styles.tsx
+++ b/src/components/Arrow/styles.tsx
@@ -27,5 +27,5 @@ export default makeStyles(
       }
     }
   }),
-  { name: 'Arrow' }
+  { name: 'Arrow', index: 1 }
 );
--- a/src/components/Builder/styles.tsx
+++ b/src/components/Builder/styles.tsx
@@ -54,5 +54,5 @@ export default makeStyles(
       color: theme.palette.common.white
     }
   }),
-  { name: 'Builder' }
+  { name: 'Builder', index: 1 }
 );
--- a/src/components/CriteriaList/styles.tsx
+++ b/src/components/CriteriaList/styles.tsx
@@ -40,5 +40,5 @@ export default makeStyles(
       fontStyle: 'italic'
     }
   }),
-  { name: 'CriteriaList' }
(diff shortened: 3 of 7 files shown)


dehall:
I can see from the deployed page that the fix worked: https://keeyanghoreshi.github.io/pathway-builder/#/ (though to be frank it's not obvious from the diffs or commits on the fork that the changes are the same as what was applied here)

That said though, this feels like a workaround to a bug rather than a fix. From some quick searching this `index: 1` approach does seem to be a common recommendation to the issue:
https://stackoverflow.com/questions/62473898/material-ui-rendering-bugs-in-production-build
https://github.com/mui-org/material-ui/issues/16609

So if this is really the only option, ok. But before we merge can we confirm - we're sure we're doing everything the right way?
https://material-ui.com/getting-started/faq/#why-arent-my-components-rendering-correctly-in-production-builds

KeeyanGhoreshi:
from what I've read, the underlying issue is that the built out CSS file gets put in under the the link to the MUI styles.  Some solutions I've seen suggest having the build process specifically put the built CSS file first, before the MUI link, but I don't really see much of a difference between this and modifying the index value.  It's an option on `makeStyles` for seemingly this exact reason, modifying the priority of the CSS files.  

I don't see an issue with using the index option but I can look into other solutions if we want a solution that doesn't involve tossing in an `index: 1` every time we make new styles..

PR #28, opened 2020-07-14
Why Did You Render fixes

blangley28:
Optimization task to fix wdyr issues. Anything related to Graph or GraphMemo has been deferred to [PATHWAYS-265](https://jira.mitre.org/browse/PATHWAYS-265). 

Marking as WIP because there is still one issue with OutlinedDiv when constructing the branch transition criteria and some styling problems related to the Graph component.

### DIFF
--- a/src/components/Graph/Graph.tsx
+++ b/src/components/Graph/Graph.tsx
@@ -4,7 +4,7 @@ import { useParams, useHistory } from 'react-router-dom';
 import graphLayout from 'visualization/layout';
 import Node from 'components/Node';
 import Arrow from 'components/Arrow';
-import { useCriteriaContext } from 'components/CriteriaProvider';
+import { useBuildCriteriaContext } from 'components/BuildCriteriaProvider';
 import { Pathway, PathwayNode } from 'pathways-model';
 import { Layout, NodeDimensions, NodeCoordinates, Edges } from 'graph-model';
 import styles from './Graph.module.scss';
@@ -174,7 +174,7 @@ const GraphMemo: FC<GraphMemoProps> = memo(function GraphMemo({
 }) {
   const { id: pathwayId } = useParams();
   const history = useHistory();
-  const { updateBuildCriteriaNodeId } = useCriteriaContext();
+  const { updateBuildCriteriaNodeId } = useBuildCriteriaContext();
   const redirectToNode = useCallback(
     nodeId => {
       const url = `/builder/${encodeURIComponent(pathwayId)}/node/${encodeURIComponent(nodeId)}`;
--- a/src/components/Graph/Graph.tsx
+++ b/src/components/Graph/Graph.tsx
@@ -1,13 +1,4 @@
-import React, {
-  FC,
-  RefObject,
-  useCallback,
-  useEffect,
-  useMemo,
-  useRef,
-  useState,
-  memo
-} from 'react';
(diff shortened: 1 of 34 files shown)


src/components/Sidebar/SidebarButton.tsx
```diff
@@ -19,7 +21,7 @@ const SidebarButton: FC<SidebarButtonProps> = ({ buttonName, buttonIcon, buttonT
         className={styles.sidebarButton}
         variant="contained"
         color="primary"
-        startIcon={buttonIcon}
+        startIcon={<FontAwesomeIcon icon={buttonIcon} />}
```
danlee1025, line 24: Does this break the never pass around JSX as a prop rule?

src/components/Graph/Graph.tsx
```diff
@@ -1,184 +1,149 @@
...
-            0
-          );
-
-          nodeDimensions[key] = { width, height };
-        });
-      }
-
-      return graphLayout(pathway, nodeDimensions);
-    }, [pathway]);
+const Graph: FC<GraphProps> = memo(function Graph({ graphContainerWidth, interactive = true }) {
```
danlee1025, line 24: Do we need the `function` declaration here? Will this not work as just an arrow function?

danlee1025:
This looks really good to me, just some very small comments. I looked through most of the code and tried to break the app and everything seems to work fine to me.

src/components/Sidebar/SidebarButton.tsx
blangley28: Need to look into this. In short yes it does but `Button` is a material component which requires a `ReactNode` as the `startIcon`. You can see the type of `buttonIcon` changed. It used to be a reactNode but now it is just the `IconDefinition`. Since we need to pass in the `ReactNode` to this material component there is nothing we can do. However, we can assume the material components are all optimized so we don't need to worry about it. The important part is that the `SidebarButton` component no longer takes in a `ReactNode` as a prop

src/components/Graph/Graph.tsx
blangley28: Yes - without the function wdyr shows this component as `undefined`. This doesn't happen for the other components because they are defined as `const Component: FC = ({}) => {};  export const memo(Component);`. But since this component is memoized here I added a name to the function here

danlee1025:
I'm good to approve after a rebase

src/components/PathwayProvider.tsx
```diff
@@ -27,34 +27,44 @@ interface PathwayProviderProps {
   children: ReactNode;
 }
 
-export const PathwayProvider: FC<PathwayProviderProps> = memo(({ children }) => {
-  const [pathways, setPathways] = useState<Pathway[]>([]);
+export const PathwayProvider: FC<PathwayProviderProps> = memo(function PathwayProvider({
```
dehall, line 30: Now that there's also a `CurrentPathwayProvider` the meaning of the name `PathwayProvider` becomes less clear. Can this be renamed to the plural Pathway**s**Provider to make the distinction more obvious? I'm also ok with making that a separate task/PR to reduce the number of changes here.

blangley28: Let me look into it after to burndown - hopefully its just a matter of changing all the names which is easy

blangley28: Renamed easily

PR #29, opened 2020-07-17
window resize

KeeyanGhoreshi:
This is to fix a bug with the window resizing, but the graph/sidebar not updating their height.  I added an event listener to take care of that, since the height update occured in a `useEffect` and wouldn't occur without the component being rerendered.


danlee1025:
Looks good to me, needs a rebase and I'll approve and merge

PR #30, opened 2020-07-17
Fix Graph Labels

mdube7:
This is to fix two issues with displaying the label on the arrows in the graph:
- The label of the arrow going to the furthest right node was being truncated after a few characters
- Labels could overlap if they were especially long

### DIFF
--- a/src/components/Arrow/Arrow.tsx
+++ b/src/components/Arrow/Arrow.tsx
@@ -5,6 +5,7 @@ import { isBranchNode } from 'utils/nodeUtils';
 import useStyles from './styles';
 import clsx from 'clsx';
 import { Tooltip } from '@material-ui/core';
+import { withStyles } from '@material-ui/core/styles';
 
 interface ArrowProps {
   edge: Edge;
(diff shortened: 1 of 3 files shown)


src/components/Arrow/Arrow.tsx
```diff
@@ -30,7 +30,7 @@ const Arrow: FC<ArrowProps> = ({ edge, edgeName, widthOffset, currentNode }) =>
       <ArrowPath points={edge.points} arrowheadId={arrowheadId} widthOffset={widthOffset} />
       {label ? (
         <text x={label.x + widthOffset} y={label.y}>
-          {label.text}
+          {label.text.length > 12 ? label.text.substring(0, 11) + '...' : label.text}
```
blangley28, line 33: This works but from a user standpoint it may be frustrating when the screen is large enough to accommodate  more than 12 characters. Perhaps make it a function of the screen size? 
Screen Shot 2020-07-17 at 2 56 15 PM

dehall: Is there a way to include the full text as a tooltip? I'm always frustrated by truncated text that doesn't display the full value when I hover over it.

mdube7: Added a tooltip for labels that are truncated.

dehall: Dang that tooltip is fresh
image

I definitely agree it would be nice to expand the text if there is space, but there are probably a lot of cases we'd need to take into account (to prevent overlapping) and that could get complicated. @blangley28 thoughts?

blangley28: Yea I think that would be really nice to have but out of scope here. Tooltip is really nice but can we make it a bit bigger? I'm not sure if it needs to be as large as the original text but its hard for me to read at first glance

src/components/Arrow/Arrow.tsx
```diff
@@ -24,14 +26,28 @@ const Arrow: FC<ArrowProps> = ({ edge, edgeName, widthOffset, currentNode }) =>
   const edgeNameNoWhitespace = edgeName.replace(' ', '');
   const arrowheadId = `arrowhead-${edgeNameNoWhitespace}`;
 
+  const CustomTooltip = withStyles({
+    tooltip: {
+      fontSize: 14
+    }
+  })(Tooltip);
+
```
blangley28, line 34: Sorry I didn't post this yesterday, was waiting for Noranda to confirm - we should add these to the MuiOverrides in `theme.tsx` since we will probably want this same styling for all tooltips

src/components/Arrow/Arrow.tsx
dehall: @blangley28 confirmed this was his last comment so this is good to merge now

PR #31, opened 2020-07-17
Delete button with confirmation popup.

Jack-Fraser16:
Added to the delete button for both the Pathway and Criteria tabs. Not super sold that the UI stuff is perfect, I think it might look nice having the text in a less wide box across multiple lines. Thoughts?


src/components/ConfirmedDeletionButton/ConfirmedDeletionButton.tsx
```diff
@@ -0,0 +1,35 @@
...
+      deleteType={deleteType}
+      deleteName={deleteName}
+      onConfirm={deleteMethod}
+    >
+      Delete
+    </ButtonPopup>
+  );
+};
+
+export default ConfirmedDeletionButton;
```
blangley28, line 35: We should memoize this
```suggestion
export default memo(ConfirmedDeletionButton);
```

src/components/CriteriaList/CriteriaTable.tsx
```diff
@@ -53,15 +54,11 @@ const CriteriaTable: FC = () => {
...
-                  size="small"
-                  startIcon={<FontAwesomeIcon icon={faTrashAlt} />}
-                  onClick={(): void => deleteCriteria(c.id)}
-                >
-                  Delete
-                </Button>
+                <ConfirmedDeletionButton
+                  deleteType="criterion"
+                  deleteName={c.label}
+                  deleteMethod={(): void => deleteCriteria(c.id)}
```
blangley28, line 60: This causes an extra re-render since the function is always the same but every time the CriteriaTable re-renders a new anonymous function is created. Make sure the function passed in a callback.
Screen Shot 2020-07-17 at 2 43 22 PM

src/components/PathwaysList/PathwaysTable.tsx
```diff
@@ -84,14 +85,11 @@ const PathwaysTable: FC = () => {
...
-                    size="small"
-                    startIcon={<FontAwesomeIcon icon={faTrashAlt} />}
-                    onClick={(): void => deletePathway(pathway.id)}
-                  >
-                    Delete
-                  </Button>
+                  <ConfirmedDeletionButton
+                    deleteType="pathway"
+                    deleteName={pathway.name}
+                    deleteMethod={(): void => deletePathway(pathway.id)}
```
blangley28, line 91: Same comment here for this function

Jack-Fraser16:
Having difficulty with figuring out why the popup is off-centered. I think its a CSS issue, but the semantic-ui pages say the pop-up should be using the position of its container to set its own relative position, and its container is the table. Going to keep playing around with it.

src/components/withConfirmationPopup/withConfirmationPopup.tsx
```diff
@@ -33,14 +47,18 @@ const withConfirmationPopup = <T extends object>(
...
 
 interface PopupContentProps {
+  deleteType: string;
+  deleteName: string;
   setOpen: Function;
   onConfirm: () => void;
 }
 
-const PopupContent: FC<PopupContentProps> = ({ setOpen, onConfirm }) => {
+const PopupContent: FC<PopupContentProps> = ({ deleteType, deleteName, setOpen, onConfirm }) => {
```
dehall, line 56: The text is nice but now this component is specific to deletion. Could we make this just one single text parameter, which is optional, that gets displayed in the popup? If not provided it would default to just "Are you sure" as before.

Jack-Fraser16:
Still trying to figure out the pop-up location bug, but am being deterred by the world of CSS.

Jack-Fraser16:
Pretty significant changes in switching to Material UI from Semantic UI. Primarily, it solved the issue of the popover on the last element of the table being skewed. The only thing to still be implement is a small arrow anchoring the popover to the delete button - a) I'm not sure we need it from a UX perspective and b) I was wondering if tooltip is the best way to go with implementing it, so any feedback on these would be great!

Jack-Fraser16:
Just talked with Dylan and I'm also going to delete the withConfirmationPopup and PathwayPopup modules before rebasing/merging because they aren't used anywhere else in the builder and both rely on semantic UI

src/components/ConfirmedActionButton/__tests__/ConfirmedActionButton.test.tsx
```diff
@@ -2,11 +2,17 @@ import React from 'react';
...
+const popupText = 'Are you sure you would like to delete the test sample?';
 
 describe('<ConfirmedActionButton />', () => {
   it('renders the popup when clicked', () => {
-    const { queryByText, getByText } = render(<ConfirmedActionButton size="large" type="accept" />);
+    const { queryByText, getByText } = render(
+      <ConfirmedActionButton
+        size="large"
+        type="accept"
+        displayText="Are you sure you would like to delete the test sample?"
```
blangley28, line 13: Is there a reason why this is copied instead of just using the value of `popupText`?

src/components/ConfirmedDeletionButton/ConfirmedDeletionButton.tsx
```diff
@@ -0,0 +1,75 @@
...
+    <div className={styles.container}>
+      <Button
+        color="secondary"
+        size="small"
+        startIcon={<FontAwesomeIcon icon={faTrashAlt} />}
+        onClick={onClickHandler}
+      >
+        Delete
+      </Button>
+      <Popover
```
blangley28, line 58: This popover is mighty beautiful but I think it should close automatically when clicking elsewhere on the screen

src/components/ConfirmedDeletionButton/styles.tsx
```diff
@@ -0,0 +1,37 @@
...
+      display: 'inline-block'
+    },
+    paper: {
+      display: 'flex',
+      flexDirection: 'row',
+      alignItems: 'center',
+      borderRadius: '10px',
+      border: '5px solid rgba(74, 74, 74, 0.5)',
+      backgroundColor: theme.palette.grey['800'],
+      color: 'white',
```
blangley28, line 15: I updated the way we handle colors in my PR. You should be using `theme.palette.common` for these colors. I think to get that to work I had to add a couple lines to `theme.tsx` though

src/components/ConfirmedActionButton/__tests__/ConfirmedActionButton.test.tsx
Jack-Fraser16: Oops no reason, but not a big deal because we're gonna delete that component anyways.

src/components/ConfirmedDeletionButton/ConfirmedDeletionButton.tsx
Jack-Fraser16: Good call.

src/components/ConfirmedDeletionButton/styles.tsx
Jack-Fraser16: What do we think about making this an item in the backlog to go through and update all colors?

blangley28: I just merged my PR so if you rebase you will have the updated palette

src/components/ConfirmedDeletionButton/ConfirmedDeletionButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+        classes={{ paper: styles.paper }}
+        open={open}
+        anchorEl={anchorEl}
+        anchorOrigin={{ vertical: 'bottom', horizontal: 'center' }}
+        transformOrigin={{ vertical: 'top', horizontal: 'center' }}
+      >
+        <div className={styles.displayText}> {displayText} </div>
+        <div className={styles.buttons}>
+          <ActionButton size="small" type="accept" onClick={(): void => onCloseHandler(true)} />
+          <ActionButton size="small" type="decline" onClick={(): void => onCloseHandler(false)} />
```
blangley28, line 71: This actually causes a wdyr issue. It doesn't show up in terminal because `ActionButton` isn't memoized. In `ActionButton.tsx` change the last line to be `export default memo(ActionButton)`. The issue is that the `onClick` function will be recreated each time. Remove the anonymous function and pass in a callback directly. You will probably need to create two methods since you won't be able to pass in the arguments

PR #32, opened 2020-07-23
Multiple Transitions + Minor Updates

blangley28:
This PR includes the following changes:

- Supporting multiple transitions for action nodes
- Updating the Node Type drop down 
- Updating the transition UI and implementing go to node (right arrow)
- Update the add node buttons (connect node not implemented)
- Fix switch node type to remove transitions 

Supporting multiple transitions for action nodes:
Screen Shot 2020-07-23 at 7 13 25 AM

Updating the Node Type drop down:
Screen Shot 2020-07-23 at 7 12 54 AM

Updated UI components:
Screen Shot 2020-07-23 at 7 11 53 AM

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -6,6 +6,7 @@
     <title>Camino Builder</title>
     <link rel="icon" href="camino-icon.png" />
     <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/semantic-ui/dist/semantic.min.css" />
+    <link href="https://fonts.googleapis.com/css2?family=Open+Sans:ital,wght@0,300;0,400;0,600;0,700;0,800;1,300;1,400;1,600;1,700;1,800&display=swap" rel="stylesheet">
     <script src="config.js" language="javascript"></script>
   </head>
   <body>
--- a/src/components/Navigation/styles.tsx
+++ b/src/components/Navigation/styles.tsx
@@ -17,7 +17,7 @@ export default makeStyles(
     },
     navigationIcons: {
       fontSize: '20px',
-      color: theme.palette.grey['800'],
+      color: theme.palette.common.gray,
       cursor: 'pointer'
     },
     pathwayName: {
--- a/src/components/Sidebar/ActionNodeEditor.tsx
+++ b/src/components/Sidebar/ActionNodeEditor.tsx
@@ -12,7 +12,6 @@ import DropDown from 'components/elements/DropDown';
 import { ActionNode, Action } from 'pathways-model';
 import { ElmLibrary } from 'elm-model';
 import useStyles from './styles';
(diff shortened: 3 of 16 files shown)


blangley28:
As well as general comments I would like to discuss:

- `nodeTypeOptions` is a list included in all 3 sidebar editors. It should probably be shared somewhere but I'm looking for feedback on where to put it.
- The "Transitions" header is included in all 3 sidebar editors because the position matters for `BranchNodeEditor`. I could move the header to `Sidebar` but then the rendering of `BranchNodeTransition`s would need to move from the `BranchNodeEditor` into `Sidebar` and I don't really like that idea either.
- I'm not sure if it just caching but sometimes the sidebar doesn't scroll. To test add a few transitions to an Observation and let me know if it scrolls correctly for you. You may need to "Empty cache and hard reload".

dehall:
There does seem to be some weirdness with scrolling. What I'm seeing is 2 "scrollable sections" -- the whole page, and the graph. So the sidebar does scroll but the graph container is included and scrolls up with it, which is not ideal. I'm seeing that on master as well so we can make that a separate task.

blangley28:
@dehall added to [PATHWAYS-287](https://jira.mitre.org/browse/PATHWAYS-287)

Jack-Fraser16:
Great job Ben! I think one thing to note with the "go to node" feature is that the only way to navigate backwards is to click on the parent node. It doesn't seem like the leftward pointing arrow on a zoomed in node is operational. Is this enough work that is warrants its own task? 
Screen Shot 2020-07-27 at 2 34 12 PM
Maybe nodeTypeOptions can be kept in sidebar.tsx and passed as a prop?

blangley28:
> I think one thing to note with the "go to node" feature is that the only way to navigate backwards is to click on the parent node. It doesn't seem like the leftward pointing arrow on a zoomed in node is operational. Is this enough work that is warrants its own task?
I didn't try to implement this. The skeleton function still exists. It looks like there is a [backlog item](https://jira.mitre.org/browse/PATHWAYS-269) for it already.

> Maybe nodeTypeOptions can be kept in sidebar.tsx and passed as a prop?
That's a possibility. But since it never changes I would prefer to not pass it as a prop. Maybe keeping it in nodeUtils and then importing in each file? I was working in the nodeUtils file today for another task and there is another static map already there.

Jack-Fraser16:
> That's a possibility. But since it never changes I would prefer to not pass it as a prop. Maybe keeping it in nodeUtils and then importing in each file? I was working in the nodeUtils file today for another task and there is another static map already there.
That sounds like a good decision for now, and then maybe clean up the nodeUtils file and consider consolidating into contexts later on?

PR #33, opened 2020-07-28
Delete Nodes and Transitions

blangley28:
Implement delete nodes and delete transitions. Modal pops up to confirm and then snackbar appears in lower right corner to confirm. Since we cannot connect nodes yet all of the transition deletions will be disabled. To test it correctly removes the transition replace the calls to `willOrphanChild` in `SidebarHeader` with `false`.

Tooltip for disabled button:
Screen Shot 2020-07-28 at 12 37 03 PM

Delete modal:
Screen Shot 2020-07-28 at 12 37 52 PM

Snackbar confirmation:
Screen Shot 2020-07-28 at 12 38 11 PM


Jack-Fraser16:
Great job Ben, the functionality is great! Some things to consider might be moving the text from the tooltip to become part of the delete modal. I think the tooltip text is a lot to read / understand for how small the text is and where it appears on the screen. It might be more user-friendly if it pops up as part of the delete modal for nodes that can't yet be deleted. That would also mean adjusting the modal to not always have the delete button, to have different text, a different icon, etc. Also, it might be redundant to have a cancel button on the modal, an X in the top right corner, and clickAway functionality - maybe get rid of the X if possible?

blangley28:
Updated the expanded node for observations and show transitions for action nodes. Implementing UNDO needs some research first ([PATHWAYS-253](https://jira.mitre.org/browse/PATHWAYS-253)

noranda:
Looks good! Issues 1 and 2 have been resolved and we have tasks for 2 and 3. One last request that I missed the first time. Can we also add a snackbar notification when we delete a transition? Otherwise this looks good to go.

PR #35, opened 2020-08-03
Connect node

Jack-Fraser16:
Connect node button implementation. Known issues include: 

1) Text underneath the Connect Node button should be higher up.
2) The tooltip will signify that there are no possible connectable nodes and the connect node button cannot be clicked. The tooltip is not rendering, likely due to issues related to the warning.
3) The options for connectable nodes doesn't always drop down, sometimes it renders in the middle or upward (in relation to the button). I think this happens when the number of options makes the dropdown larger than the size allotted for the sidebar on the screen.  
4) Label re-adjustment needs to be fine tuned when new transitions are added - to see what I mean, create a transition between any of the last three nodes on the HER2+ pathway. 

Feedback / advice / commiseration on any of these issues would be much appreciated - thanks!


src/components/ConnectNode/ConnectNode.tsx
```diff
@@ -0,0 +1,128 @@
...
+  );
+  previousNodes.push(rootNode);
+
+  for (const nodeKey of Object.keys(pathway.nodes)) {
+    const currNode = pathway.nodes[nodeKey];
+    if (!rootNode.transitions.some(transition => transition.transition === nodeKey)) {
+      if (!previousNodes.some(node => node.key === currNode.key))
+        connectableNodes.push({ label: currNode.label, value: currNode.key ?? '' });
+    }
+  }
```
blangley28, line 63: This can be cleaned up to be more readable:
```suggestion
  Object.keys(pathway.nodes).forEach((nodeKey) => {
    const node = pathway.nodes[nodeKey];
    const rootNodeConnectsToNode = rootNode.transitions.some(transition => transition.transition === nodeKey);
    const nodeIsAncestor = ancestorNodes.some(node => node.key === node.key)
    if (!rootNodeConnectsToNode && !nodeIsAncestor && node.key) 
        connectableNodes.push({ label: node.label, value: node.key});
  }
```

src/components/ConnectNode/ConnectNode.tsx
```diff
@@ -0,0 +1,128 @@
...
+        pathway,
+        rootNodeKey,
+        pathway.nodes[transition.transition].key ?? ''
+      );
+      if (previousNodes !== undefined && previousNodes.length > 0) {
+        if (!ancestors.some(node => node === currNode)) ancestors.push(currNode);
+        ancestors = ancestors.concat(previousNodes);
+      }
+    }
+  }
```
blangley28, line 40: This can clean up to be
```suggestion
 currNode.transitions.forEach((transition) => {
    if (transition.transition === rootNodeKey) ancestors.push(currNode);
    else {
      previousNodes = getAncestorNodes(pathway, rootNodeKey, transition.transition);
      if (previousNodes?.length) {
        if (!ancestors.some(node => node.key === currNode.key)) ancestors.push(currNode);
        ancestors = ancestors.concat(previousNodes);
      }
    }
  }
```

src/components/ConnectNode/ConnectNode.tsx
```diff
@@ -0,0 +1,128 @@
...
+  const [open, setOpen] = useState(false);
+  const options = getConnectableNodes(pathway, rootNode);
+  const { updatePathway } = usePathwaysContext();
+  const optionsAvailable: boolean = options.length > 0;
+
+  const connectToNode = useCallback(
+    (event: ChangeEvent<{ value: string }>): void => {
+      const nodeKey = event?.target.value;
+      const newPath = addTransition(pathway, rootNode.key ?? '', nodeKey);
+      updatePathway(newPath);
```
blangley28, line 79: We've been making this one line in other places
```suggestion
      updatePathway(addTransition(pathway, rootNode.key ?? '', nodeKey));
```

src/components/ConnectNode/ConnectNode.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  return connectableNodes;
+};
+
+const ConnectNodeButton: FC<ConnectNodeButtonProps> = ({ pathway, rootNode }) => {
+  const styles = useStyles();
+  const [open, setOpen] = useState(false);
+  const options = getConnectableNodes(pathway, rootNode);
+  const { updatePathway } = usePathwaysContext();
+  const optionsAvailable: boolean = options.length > 0;
```
blangley28, line 73: It should automatically know this is a boolean
```suggestion
  const optionsAvailable = options.length > 0;
```

src/components/ConnectNode/ConnectNode.tsx
```diff
@@ -0,0 +1,128 @@
...
+          open={!optionsAvailable}
+          placement="top"
+        >
+          <SidebarButton
+            buttonName="Connect Node"
+            buttonIcon={faLevelDownAlt}
+            buttonText="Create a transition to an existing node in the pathway."
+            onClick={(): void => {
+              optionsAvailable && setOpen(true);
+            }}
```
blangley28, line 99: This causes a wdyr make this a usecallback

```
const showDropdown = useCallback(() => {
     if (optionsAvailable) setOpen(true);
}, [optionsAvailable]);

...
<SidebarButton ... onClick={showDropdown} />
```

Jack-Fraser16:
Just re-based and implemented Ben's feedback. Going to start writing some tests in the meantime while I wait for anymore feedback.

src/components/Sidebar/ConnectNodeButton.tsx
```diff
@@ -0,0 +1,82 @@
...
+const ConnectNodeButton: FC = () => {
+  const { pathway, pathwayRef } = useCurrentPathwayContext();
+  const { currentNode, currentNodeRef } = useCurrentNodeContext();
+  const { updatePathway } = usePathwaysContext();
+
+  const styles = useStyles();
+  const [open, setOpen] = useState(false);
+
+  let options: Array<{ label: string; value: string }> = [];
+  if (currentNode && pathway) options = getConnectableNodes(pathway, currentNode);
```
blangley28, line 24: Since options is never changed lets make it `const`
```suggestion
  const options: Array<{ label: string; value: string }> = currentNode && pathway ? getConnectableNodes(pathway, currentNode) : [];
```

src/components/Sidebar/ConnectNodeButton.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const optionsAvailable = options.length > 0;
+
+  const connectToNode = useCallback(
+    (event: ChangeEvent<{ value: string }>): void => {
+      const nodeKey = event?.target.value;
+      if (pathwayRef.current && currentNodeRef.current)
+        updatePathway(addTransition(pathwayRef.current, currentNodeRef.current.key ?? '', nodeKey));
```
blangley28, line 31: Adding a transition from '' to nodeKey doesn't make sense so add it to the if statement
```suggestion
      if (pathwayRef.current && currentNodeRef.current?.key)
        updatePathway(addTransition(pathwayRef.current, currentNodeRef.current.key, nodeKey));
```

src/utils/__tests__/nodeUtils.test.ts
```diff
@@ -38,4 +38,25 @@ describe('node util methods', () => {
     const parents = findParents(samplepathway, 'ChemoMedication');
     expect(parents).toEqual(['N-test', 'Surgery']);
   });
+
+  describe('get connectable nodes', () => {
```
blangley28, line 42: For these tests - instead of returning true/false can we make them check the return objects are identical (since we can figure out what they should be from the samplePathway)

blangley28:
Also I am seeing this warning (which I cannot replicate on master). 
Steps to reproduce:
1. Open up the Sample Pathway for Testing
2. Add a new node

Screen Shot 2020-08-05 at 8 21 32 AM

src/utils/__tests__/nodeUtils.test.ts
Jack-Fraser16: Yes. I'm going to change the second one test to check the return object, but keep the first one as is because it's the only way I can think of to quickly check that every node in the pathway is finding potential transitions as expected.

Jack-Fraser16:
Still figuring out the forwardRef error.

src/utils/nodeUtils.ts
```diff
@@ -70,3 +70,47 @@ export function redirect(
...
+  const ancestorNodes = getAncestorNodes(pathway, rootNode.key ?? '', 'Start');
+  ancestorNodes.push(rootNode);
+
+  Object.keys(pathway.nodes).forEach(nodeKey => {
+    const node = pathway.nodes[nodeKey];
+    const rootNodeConnectsToNode = rootNode.transitions.some(
+      transition => transition.transition === nodeKey
+    );
+    const nodeIsAncestor = ancestorNodes.some(node => node.key === nodeKey);
+    if (!rootNodeConnectsToNode && !nodeIsAncestor && node.key)
```
noranda, line 111: It looks like we aren't allowing transitions to ancestor nodes. Is this what we want? This prevents loops, but I wasn't sure if we wanted to prevent those.

src/components/Sidebar/ConnectNodeButton.tsx
```diff
@@ -0,0 +1,82 @@
...
+  );
+
+  const showDropdown = useCallback(() => {
+    if (optionsAvailable) setOpen(true);
+  }, [optionsAvailable]);
+
+  return (
+    <div>
+      {!open && (
+        <Tooltip
```
noranda, line 44: We want to move this tooltip into the `SidebarButton` component otherwise the hover area is the entire div, not just the button. Moving it also means we don't need to pass in a ref, which is where your ref error is coming from.

To do so, remove the `Tooltip` wrapper here and replace the `SidebarButton` with:

```
<SidebarButton
  buttonName="Connect Node"
  buttonIcon={faLevelDownAlt}
  buttonText="Create a transition to an existing node in the pathway."
  onClick={showDropdown}
  disabled={!optionsAvailable}
  hasTooltip={!optionsAvailable}
  tooltipTitle="There are no possible nodes to connect to."
/>
```
We also want this button to be disabled, so we pass that prop in as well.

Then, we can refactor the `SidebarButton` component as follows:

```
import React, { FC, memo } from 'react';
import { Button, Tooltip } from '@material-ui/core';
import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
import { IconDefinition } from '@fortawesome/fontawesome-svg-core';

interface SidebarButtonProps {
  buttonName: string;
  buttonIcon: IconDefinition;
  buttonText: string;
  extraMargin?: boolean;
  onClick?: () => void;
  hasTooltip?: boolean;
  tooltipTitle?: string;
  disabled?: boolean;
}

const SidebarButton: FC<SidebarButtonProps> = ({
  buttonName,
  buttonIcon,
  buttonText,
  extraMargin = false,
  onClick,
  hasTooltip = false,
  tooltipTitle = '',
  disabled = false
}) => {
  const styles = useStyles();

  const ConditionalTooltip: FC = ({ children }) =>
    hasTooltip ? (
      <Tooltip title={tooltipTitle} placement="top">
        <span>{children}</span>
      </Tooltip>
    ) : (
      <>{children}</>
    );

  return (
    <div className={extraMargin ? styles.sidebarButtonGroupExtraMargin : styles.sidebarButtonGroup}>
      <ConditionalTooltip>
        <Button
          className={styles.sidebarButton}
          variant="contained"
          color="primary"
          startIcon={<FontAwesomeIcon icon={buttonIcon} />}
          onClick={onClick}
          disabled={disabled}
        >
          {buttonName}
        </Button>
      </ConditionalTooltip>
      <div className={styles.sidebarButtonText}>{buttonText}</div>
    </div>
  );
};

export default memo(SidebarButton);
```

Here we use the `hasTooltip` flag to conditionally show the tooltip just on the button. This also allows us to add tooltips in the future on any sidebar button we like.

The last thing we'd need to do is update the disabled styles in the `styles/theme.tsx` file on line 76 as follows:

```
const materialUiOverridesBase = {
  MuiButton: {
    root: {
      borderRadius: 0,
      '&$disabled': {
        backgroundColor: colors.grayMedium,
        color: colors.white
      }
    },
    contained: {
      '&$disabled': {
        backgroundColor: colors.grayMedium,
        color: colors.white
      }
    },
    ...
```

src/components/Sidebar/ConnectNodeButton.tsx
```diff
@@ -0,0 +1,82 @@
...
+      {open && optionsAvailable && (
+        <div className={styles.connectDropdown}>
+          <DropDown
+            id="transitions"
+            label="Node To Connect To"
+            options={options}
+            onChange={connectToNode}
+          />
+          <div className={styles.connectText}>
+            Select node from list or click node on the right to add transition.
```
noranda, line 66: Since we don't yet have the click node feature working, let's change this for now.
```suggestion
            Select node from list to add transition.
```

src/components/Sidebar/ConnectNodeButton.tsx
```diff
@@ -0,0 +1,82 @@
...
+            <Button
+              className={styles.cancelButtonDropdown}
+              size="small"
+              variant="text"
+              onClick={(): void => setOpen(false)}
```
KeeyanGhoreshi, line 71: I think the `showDropdown` callback should be reworked so that it can be used in both this `onClick` and the one in the sidebar button.  Perhaps a toggle, or just have it accept a parameter.

src/utils/nodeUtils.ts
Jack-Fraser16: I was also unsure about this. I think it should be a different task to research/make a decision about this because I see how loops would be useful, but loops are also only useful if the clinician can tell that a patient is on the loop for the x time - otherwise it would just appear that they're in that node on the pathway for the first time. Currently we don't store the information to figure out if someone is in a loop / what # cycle in the loop so there is no way to display it, meaning the loop wouldn't give us any information that undoing or advancing a patient back and forth across two nodes would.

noranda:
Looks good! Just needs a rebase. I also added a task in JIRA to deal with the duplicate node name issue.

PR #36, opened 2020-08-05
Update Highlights

mdube7:
Arrows to children are always red when the parent node is selected.
Children of observations are now blue instead of red.

### DIFF
--- a/src/components/Arrow/Arrow.tsx
+++ b/src/components/Arrow/Arrow.tsx
@@ -1,7 +1,6 @@
 import React, { FC } from 'react';
 import { Edge, Coordinate } from 'graph-model';
 import { PathwayNode } from 'pathways-model';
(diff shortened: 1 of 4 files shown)


src/components/Arrow/Arrow.tsx
```diff
@@ -21,7 +20,7 @@ interface ArrowPathProps {
 
 const Arrow: FC<ArrowProps> = ({ edge, edgeName, widthOffset, currentNode }) => {
   const styles = useStyles();
-  const isCurrentBranchArrow = isBranchNode(currentNode) && edge.start === currentNode.key;
+  const isCurrentBranchArrow = edge.start === currentNode.key;
```
blangley28, line 23: Since we use Branch exclusively with BranchNodes but ActionNodes can also have multiple arrows I think we should rename this variable to just `isCurrentArrow`

PR #37, opened 2020-08-11
Disabled criteria button when no criteria options exist.

Jack-Fraser16:


Jack-Fraser16:
Tried to implement an alternative select criteria button, but this seemed to be the most straightforward solution.

Jack-Fraser16:
Additionally, to test, go to the criteria tab and delete both default criteria.

src/components/Sidebar/BranchTransition.tsx
```diff
@@ -141,6 +143,9 @@ const BranchTransition: FC<BranchTransitionProps> = ({ transition }) => {
           buttonIcon={faThList}
           buttonText="Add previously built or imported criteria logic to branch node."
           onClick={handleUseCriteria}
+          disabled={!criteriaAvailable}
+          hasTooltip={!criteriaAvailable}
+          tooltipTitle="There are no criteria options available."
```
mdube7, line 148: This might be a @noranda question but do we want to inform the user where/how they can load criteria in this tooltip or is that overkill? Maybe something along the lines of "Criteria list is empty. Import criteria to enable this button"

Jack-Fraser16: I think that's a great call.

noranda: I wouldn't say "Import criteria", since eventually we will be able to import or build criteria from the list. Instead, let's change it to: "No criteria to select. Add criteria to list to use."

PR #40, opened 2020-08-14
Minimal criteria builder

danlee1025:
Added `CriteriaBuilder` component
- hardcoded to support "Age Range" and "Gender"
- generates cql string that is added to transition when the criteria is saved
- Styling is similar to CDS-A-T but most of the functionality for element selection is there

Drawbacks:
- Cannot edit a criteria after it has been saved. We currently can't tell the difference between a criteria that we've built and a criteria that was imported. This should change after the criteria model update.
- All built criteria is automatically added to criteria list for now.


dehall:
Looks good! One thing I notice is that if you collapse the sidebar while you are building the criteria, it seems to de-sync, so when you re-open the sidebar it loses the "currently building" state. Though it seems to reconnect when you change something on the builder side

mdube7:
1. When you are building the criteria if you click the button in the top right to view the pathway and go back you lose progress. 
2. If you have an observation with 1 node and click build criteria the sidebar shows that you are currently building criteria. There seems to be an issue if there are 2+ nodes and you click build criteria on any of them. The sidebar does not show that the criteria it is being built and this puts the build criteria into a state where you cannot cancel or complete building that criteria.
image

src/components/CriteriaBuilder/CriteriaBuilder.tsx
```diff
@@ -0,0 +1,162 @@
...
+    setMinimumAge(0);
+    setMaximumAge(0);
+  }, []);
+
+  const genderString = `The patient's gender is ${gender}`;
+  const ageRangeString = `The patient's age is between ${minimumAge} years and ${maximumAge} years`;
+
+  useEffect(() => {
+    const cql = `AgeInYears() >= ${minimumAge} and AgeInYears() < ${maximumAge}`;
+    if (minimumAge >= 0 && maximumAge > 0) {
```
mdube7, line 75: Do we need to ensure that minimumAge <= maximumAge as well?

danlee1025: addressed

danlee1025:
@dehall @mdube7 all comments should be addressed now!

dehall:
I just tried changing one criteria in the HER2+ pathway to "gender = male" and I'm getting a result I wouldn't have expected. Relevant section is below, full version is attached: 
[Early Stage HER2+ (Draft) (3).json.txt](https://github.com/mcode/pathway-builder/files/5098231/Early.Stage.HER2%2B.Draft.3.json.txt)


```
    "NodeStatus": {
      "label": "Node Status",
      "transitions": [
        {
          "transition": "TumorSize",
          "condition": {
            "description": "male_criteria_display",
            "cql": "dqvhWd6BND"
          }
        },
```

Specifically that `"cql": "dqvhWd6BND"` is a little strange and doesn't seem to reference anything else. I would have expected something like `Patient.gender.value = 'male'` here. I haven't dug closely into the code to figure out where this is getting written. That said though I do see the ELM contains relevant pieces for gender=male

dehall:
One last thing - this will need to be rebased since the "export to CPG" feature just got merged in. Can you check if the CQL generated by this builder gets put in the right place when a pathway is exported to CPG? It would be in an applicability statement for an action on a PlanDefinition (or, just ping me when you rebase and I'll check) I expect it will work correctly without further changes but want to confirm

PR #41, opened 2020-08-20
Refactor the Graph

noranda:
This PR removes the old Graph implementation, replacing it with a new `DagreGraph` set of components that provides a cleaner and more efficient integration with the Dagre graph library.

Previously, anytime we made a change, we were spinning up the graph, setting the nodes, then computing the heights, widths, and coordinates for ALL the nodes, whether or not they changed, then rendering. With the refactor, each node is in charge of rendering itself, so individual nodes can notify the graph when they change without having to render the entire graph. It will now only perform calculations for nodes that change. This significantly speeds up the rendering of the graph. It also provides a cleaner implementation which is easier to read and understand.

Given the size of the PR and the changes made, I recommend 3 reviewers to try to find any edge cases that I may have missed.

## Builder CSS refactor

The `Builder` component has been updated to use flexbox to size itself to take up the full browser viewport. This was previously done with JS. By switching to flexbox, we can remove the header reference tracking as well as the use effects that resize the builder as the browser window resizes, greatly simplifying the component!

## The DagreGraph

The `DagreGraph` is made up of a several components:

### `GraphProvider`

This provides the layout integration with Dagre. It's made up of three contexts:

#### GraphContext

The `GraphContext` provides two values, `graph` and `reflow`.

- `graph` is the Dagre graphlib instance. Nodes will register themselves with it along with their transitions.
- `reflow` is a function that, when called, will recalculate the positions of the nodes and transition lines and labels.

#### SetGraphWidthContext

The `SetGraphWidthContext` provides a function that sets the viewport width of the graph. This is used by the `GraphRoot` component.

#### GraphCoordinatesContext

The `GraphCoordinatesContext` provides two values, `nodes` and `edges`. These objects contain all of the positions and sizes for the nodes and transitions in the graph, taken from Dagre.

### `GraphRoot`

The `GraphRoot` component is the relatively positioned `div` that the nodes and transition arrows are drawn inside of. It uses the coordinates to calculate the max width and height needed to draw them, and adds a little bit of a buffer to that.

It is also in charge of keeping track of the viewport size and resizing the graph when that changes.

### `GraphNodes`

The `GraphNodes` component loops over all of the pathway nodes and renders each one. This component is the one that re-renders every time a node changes, so it is small, simple, and fast.

### `Node`

The `Node` component is responsible for rendering each node, as well as keeping track of its width and height. When it is mounted it registers itself with the graph, and when it is remove it unregisters itself. It also keeps track of its transitions, updating the Dagre graph in a similar manner when changes are detected.

It makes use of two helper components, `NodeIcon` which renders the icon next to the node name and `ExpandedNode` which renders the content inside the node when expanded. 

### `GraphEdges`

The `GraphEdges` component is responsible for rendering the SVG paths and labels that make up the pathway transitions.

I was able to refactor the old `Arrow` component so that all of the `<path />` and `<text />` tags render inside of one `svg`, removing the need to duplicate the arrowhead in every line.


dehall:
Another minor note - typing into the "Criteria Display" field on a transition on a branch node is super laggy.
I know we noticed this before but it seems even worse now.  I think I tried all the other text fields and that's the only one that really stands out.

noranda:
@dehall Good catch on the flashing node. The reason it was flashing is because Dagre needs it to render first to calculate its dimensions before rendering it in its correct spot. I fixed it to render off screen so you shouldn't see the flashing anymore.

Regarding the lag when typing in the Criteria Display field, have you tested this on production mode? React does all kinds of things in development mode that slow it down for debugging purposes. I recommend only testing responsiveness in production mode. To do so:

1. Temporarily delete the `homepage` attribute in `package.json` (line 5)
2. Run `yarn build`
3. Go into the build directory (`cd build`) and run `python -m SimpleHTTPServer $1` (or use any other static server command)
4. Open app in port listed (ex. `http://localhost:8000/`)

dehall:
Oh yeah it is a lot faster in production mode. No lag at all. Strange. 

So that was all the functional issues. With those resolved this looks awesome. I haven't gone super deep into the code changes yet, but that's been the theme of this whole year sadly. I'll keep looking but will leave the 2 approvals for team members who have more experience with the specifics.

src/components/DagreGraph/useCoordinatesCalculator.ts
```diff
@@ -0,0 +1,102 @@
...
+export interface GraphCoordinates {
+  nodes: NodeCoordinates;
+  edges: Edges;
+}
+
+export interface NodeCoordinates {
+  [nodeName: string]: Node;
+}
+
+interface ExtractCoordinatesInterface {
```
danlee1025, line 14: I don't think this interface is being used anywhere, do we want to remove it?

danlee1025:
It looks like the sidebar's width is bigger than it is on master, don't think it's a big deal either way, just wondering if this was intentional.

src/components/DagreGraph/useCoordinatesCalculator.ts
noranda: Good catch! Will delete.

noranda:
> It looks like the sidebar's width is bigger than it is on master, don't think it's a big deal either way, just wondering if this was intentional.

This was likely a result of switching to use flexbox to handle calculating the sidebar's dimensions, which means we don't need to rely on JS to calculate them for us. It may also have to do with fixing the bug where the collapse button wasn't showing up on the Start node. Either way, as it looks like only a 4px difference, I'm okay with it. 👍

src/components/DagreGraph/GraphRoot.tsx
```diff
@@ -0,0 +1,63 @@
...
+}
+
+const GraphRoot: FC<GraphRootProps> = ({ children }) => {
+  const styles = useStyles();
+  const graphRef = useRef<HTMLDivElement>(null);
+  const setGraphWidth = useSetGraphWidth();
+  const { nodes: nodeCoordinates } = useGraphCoordinates();
+  const graphWidth = useMemo(() => {
+    return Math.ceil(
+      150 +
```
blangley28, line 18: Where is this 150 coming from? Is this just a padding?

noranda: Yep, this is just padding so that the graph doesn't rest right up against the edge.

noranda: I've moved the number into a `GRAPH_PADDING` variable to make it easier to see what it's used for

noranda:
> Played around with everything I could think of with the graph and looks good. I noticed however nodes don't automatically expand anymore when you start adding details to them. I personally liked that feature - was that a conscious change?

Nope, great catch! This was something I missed when doing the refactor. I've pushed a fix that restores the behavior

PR #42, opened 2020-08-21
CPG Multiple Conditions

blangley28:
The previous PR was missing applicability conditions for a chain of branch nodes. Updated to recursively check each parent to ensure all the conditions are included in the node. This works but there are still a few issues:
1. Right now duplicates are added (ie. the same condition may appear more than once)
2. This does not handle and/or correctly. Take the following 
Screen Shot 2020-08-21 at 6 40 56 AM
For the `Chemotherapy TCH AC + TH` node the two conditions should be `N+` OR `N0 && T > 2cm`. However the code currently just pushes all of these applicability statements to the condition. According to the [PlanDefinition spec](https://www.hl7.org/fhir/plandefinition-definitions.html#PlanDefinition.action.condition), multiple applicability conditions means a logical AND. This is a problem since we want OR. I am unsure how ORs are supposed to be represented in PlanDefinitions but I can post in Zulip or created a Jira ticket for the CPG IG. It will be difficult (and annoying) if we have to say `N+ || N0` AND `N+ || T > 2cm`.

### DIFF
--- a/src/utils/__tests__/cpg.test.ts
+++ b/src/utils/__tests__/cpg.test.ts
@@ -1,5 +1,6 @@
 import samplepathway from './fixtures/sample_pathway.json';
-import { toCPG, createActivityDefinition, createPlanDefinition } from 'utils/cpg';
+import { toCPG, createActivityDefinition, createPlanDefinition, getExpressions } from 'utils/cpg';
+import { PlanDefinition } from 'fhir-objects';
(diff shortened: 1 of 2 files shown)


blangley28:
This is OBE (see [PATHWAYS-352](https://jira.mitre.org/browse/PATHWAYS-352)

PR #43, opened 2020-08-21
Support multiple statements in ELM criteria import

blangley28:
Importing an ELM file for criteria all statements (minus the default ones which were already excluded) are imported. The underlying Elm library is the same for both criteria but the statement is different. This statement is used to populate the `transition.condition.cql` while the elm library is used for the `transition.condition.elm`. [Test JSON file](https://jira.mitre.org/secure/attachment/42397/Node-Size.json) which has two statements (N0 and N+) for testing - when importing the file you will see both statements appear in the criteria list

### DIFF
--- a/src/components/CriteriaProvider.tsx
+++ b/src/components/CriteriaProvider.tsx
@@ -92,7 +92,6 @@ export const CriteriaProvider: FC<CriteriaProviderProps> = memo(({ children }) =
       if (event.target?.result) {
         const rawElm = event.target.result as string;
         const newCriteria = jsonToCriteria(rawElm);
-        console.log(newCriteria);
         if (newCriteria) setCriteria(currentCriteria => [...currentCriteria, ...newCriteria]);
       } else alert('Unable to read that file');
     };
--- a/src/components/CriteriaProvider.tsx
+++ b/src/components/CriteriaProvider.tsx
@@ -19,7 +19,7 @@ interface CriteriaContextInterface {
   criteria: Criteria[];
   addCriteria: (file: File) => void;
   deleteCriteria: (id: string) => void;
-  addElmCriteria: (elm: ElmLibrary, criteriaName: string) => string;
+  addElmCriteria: (elm: ElmLibrary) => Criteria[];
 }
 
 export const CriteriaContext = createContext<CriteriaContextInterface>(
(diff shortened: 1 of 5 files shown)


src/utils/builder.ts
```diff
@@ -386,21 +387,18 @@ export function setTransitionCondition(
...
     );
 
-    const cql = criteriaLabel ? criteriaLabel : getElmStatement(elm).name;
-
     if (foundTransition)
       foundTransition.condition = {
         description: description,
-        cql: cql,
-        elm: elm
+        cql: criteria.statement,
```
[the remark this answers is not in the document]
blangley28: Rebased - asked Dan to come take a look to make sure it all works well (I tested myself and it looks like it does)

src/components/CriteriaProvider.tsx
```diff
@@ -89,7 +92,8 @@ export const CriteriaProvider: FC<CriteriaProviderProps> = memo(({ children }) =
       if (event.target?.result) {
         const rawElm = event.target.result as string;
         const newCriteria = jsonToCriteria(rawElm);
-        if (newCriteria) setCriteria(currentCriteria => [...currentCriteria, newCriteria]);
+        console.log(newCriteria);
```
danlee1025, line 95: I think this was left here during debugging.

src/utils/builder.ts
```diff
@@ -381,23 +382,20 @@ export function setTransitionCondition(
...
 
-    if (foundTransition) {
+    if (foundTransition)
       foundTransition.condition = {
-        description,
-        elm,
-        criteriaSource: criteriaSource ?? getElmStatement(elm).name,
-        cql: cql ?? ''
+        description: description,
+        cql: criteria.statement,
```
danlee1025, line 395: When I export the pathway it looks like the cql generated by the criteria builder is no longer included in the criteria. I'm assuming it has to do with some of the changes in this function. Did we still want to include the raw cql in the pathway?

blangley28: From previous discussions we decided the cql should just be the statement name and the raw library exported separately (which right now is just in ELM). @dehall is this still the plan with current developments?

src/components/CriteriaProvider.tsx
blangley28: Good catch - removed.

src/utils/builder.ts
dehall: So there are 2 options - include the CQL statement directly in this field, or include a name for the statement in this field and the CQL itself in an exported library. There are probably advantages to both, but at this point I don't have any strong preference as long as the statement goes somewhere. We do want to include the CQL as well as ELM because cqf-ruler can only execute CQL, not ELM.

danlee1025: Ok I just wanted to make sure this wasn't a rebase error. I think the way Ben has it now should be fine then.

dehall: Yeah so we will eventually need something to export the CQL in the CPG, but that's not part of this task. Right now it's exported as ELM which works but cqf-ruler can't parse that

PR #44, opened 2020-08-28
add new fields to criteria model

KeeyanGhoreshi:
The basic idea to change the criteria model is accounted for, but a lot of the code ended not needing a change since we're assuming that ELM/CQL is always available.  A lot of the code happily functions without change even with the model altered to make the ELM and associated statement optional.  I changed some code so that the criteria builder will produce builder model criteria, which with an associated required CQL field, didn't require too many changes.  There would be some additional changes required if the builder model lost the (required) CQL field.

Additionally, there's no real way to import CQL files directly.  I presume our preferred method would be to just convert CQL to ELM on import and then just deal with the elm normally.

You can test this branch by ensuring that you can add/create/import criteria, and reuse saved builder criteria.


src/types/criteria-model.d.ts
```diff
@@ -4,9 +4,28 @@ declare module 'criteria-model' {
...
     label: string;
-    version: string;
+    version?: string;
     modified: number;
-    elm: ElmLibrary;
-    statement: string;
+    elm?: ElmLibrary;
+    statement?: string;
+    cql?: string;
+    builder?: BuilderModel;
```
blangley28, line 12: If this is too complex it's probably not worth it but I think the desired functionality is that at least one of elm, cql, or builder must be set. I found [this stack overflow](https://stackoverflow.com/a/60617060) answer which might be helpful.
Also - I think statement should be required because each type of criteria should have it

src/utils/builder.ts
```diff
@@ -389,13 +389,15 @@ export function setTransitionCondition(
       (transition: Transition) => transition.id === transitionId
     );
 
-    if (foundTransition)
+    if (foundTransition) {
+      const cql = criteria.statement || criteria.builder?.cql || '';
```
blangley28, line 393: See my comment above about statement but I think the builder type should know its statement. Placing the entire CQL here is outdated and being updated in [PATHWAYS-343](https://jira.mitre.org/browse/PATHWAYS-343). I'll let you and Dan figure out who should update it but this should be a boolean valued expression (such as the statement)

src/components/CriteriaProvider.tsx
```diff
@@ -61,13 +69,15 @@ function elmLibraryToCriteria(elm: ElmLibrary, custom = false): Criteria[] {
...
 }
 
-function jsonToCriteria(rawElm: string): Criteria[] | undefined {
-  const elm = JSON.parse(rawElm);
-  if (!elm.library?.identifier) {
+function jsonToCriteria(rawCriteria: string): Criteria[] | undefined {
+  const criteria = JSON.parse(rawCriteria);
+  if (criteria.type) {
+    builderModelToCriteria(criteria);
+  } else if (!criteria.library?.identifier) {
```
blangley28, line 76: Did you intend to do something with this value? The result is not stored anywhere so the elmLibraryToCriteria function will be the result returned

src/components/CurrentCriteriaProvider.tsx
```diff
@@ -35,16 +37,16 @@ interface CurrentCriteriaProviderProps {
 
 export const CurrentCriteriaProvider: FC<CurrentCriteriaProviderProps> = memo(({ children }) => {
   const [currentCriteriaNodeId, setCurrentCriteriaNodeId] = useState<string>('');
-  const [currentCriteriaCql, setCurrentCriteriaCql] = useState<CurrentCriteriaCql | null>(null);
+  const [currentCriteriaCql, setCurrentCriteria] = useState<BuilderModel | null>(null);
```
danlee1025, line 40: I think you forgot to rename `currentCriteriaCql` -> `currentCriteria` here

src/utils/builder.ts
```diff
@@ -389,13 +389,15 @@ export function setTransitionCondition(
       (transition: Transition) => transition.id === transitionId
     );
 
-    if (foundTransition)
+    if (foundTransition) {
+      const cql = criteria.statement;
       foundTransition.condition = {
         description: description,
-        cql: criteria.statement,
+        cql: cql,
```
danlee1025, line 396: I think you can just revert this to how it was before, `cql: criteria.statement` so you don't need to set the unnecessary cql const on line 393.

src/components/CriteriaProvider.tsx
```diff
@@ -61,13 +70,15 @@ function elmLibraryToCriteria(elm: ElmLibrary, custom = false): Criteria[] {
   });
 }
 
-function jsonToCriteria(rawElm: string): Criteria[] | undefined {
-  const elm = JSON.parse(rawElm);
-  if (!elm.library?.identifier) {
+function jsonToCriteria(rawCriteria: string): Criteria[] | undefined {
+  const criteria = JSON.parse(rawCriteria);
+  if (criteria.type) {
+    return [builderModelToCriteria(criteria)];
```
dehall, line 76: Does this branch ever get called? Seems like this function is only used when uploading a file, and I don't expect we're going to upload the builder model directly

KeeyanGhoreshi: you're right, builderModelToCriteria wouldn't handle the raw string anyway

blangley28: FYI lint is throwing an error here now become `builderModelToCriteria` is imported but never used

danlee1025: Have to remove the function definition since it's not being used anymore and causing a lint error.

src/components/CriteriaProvider.tsx
```diff
@@ -61,13 +70,13 @@ function elmLibraryToCriteria(elm: ElmLibrary, custom = false): Criteria[] {
...
-  if (!elm.library?.identifier) {
+function jsonToCriteria(rawCriteria: string): Criteria[] | undefined {
+  const criteria = JSON.parse(rawCriteria);
+  if (!criteria.library?.identifier) {
     alert('Please upload ELM file');
     return;
   }
-  return elmLibraryToCriteria(elm);
+  return elmLibraryToCriteria(criteria);
 }
```
blangley28, line 80: So since this only has one branch now and the `elmLibraryToCriteria` requires the argument to be an `ElmLibrary` can we revert these changes back to the names `elm` instead of `criteria`?

src/utils/builder.ts
```diff
@@ -510,7 +511,7 @@ export function setTransitionConditionElm(
 
     if (foundTransition?.condition) {
       foundTransition.condition.elm = criteria.elm;
-      foundTransition.condition.cql = criteria.statement;
+      foundTransition.condition.cql = criteria.statement || criteria.builder?.cql || '';
```
blangley28, line 514: `Criteria.statement` is required now so this change can be removed

blangley28 (approved):

PR #45, opened 2020-09-01
Update Pathways Model

blangley28:
While working on the CPG updates I ran into issues with the data model so I decided to pull [PATHWAYS-338](https://jira.mitre.org/browse/PATHWAYS-338) in. I updated the following in the data model:

- `Pathway.description` is now required
- `Precondition.id` is now required
- `PathwayNode.key` is now required
- `ActionNode.action` is now a single item instead of a list
- `Action.id` is now required
- `Transition.id` is now required

I still need to figure out what to do for the export (hence the mergeElm lint error) which strips these ids away but its in a good enough state for people to start looking at to make sure I didn't miss anything.

### DIFF
--- a/public/static/pathways/her2_pathway.json
+++ b/public/static/pathways/her2_pathway.json
@@ -181,16 +181,27 @@
         }
       ]
     },
-    "NoEndocrineTherapy": {
-      "key": "NoEndocrineTherapy",
-      "label": "No Endocrine Therapy",
-      "action": {},
-      "transitions": []
-    },
     "EndocrineTherapy": {
       "key": "EndocrineTherapy",
       "label": "Endocrine Therapy",
-      "action": {},
+      "action": {
+         "id": "1",
+         "type": "create",
+         "description": "Hormone Therapy",
+         "resource": {
+           "resourceType": "ServiceRequest",
+           "code": {
+             "coding": [
+               {
+                 "system": "http://snomed.info/sct",
+                 "code": "169413002",
+                 "display": "Hormone therapy (procedure)"
+               }
+             ],
+             "text": "Hormone therapy (procedure)"
+           }
+         }
+      },
       "transitions": []
     }
   },
--- a/public/static/pathways/her2_pathway.json
+++ b/public/static/pathways/her2_pathway.json
@@ -75,8 +75,7 @@
     "ChemotherapyTHWeekly": {
       "key": "ChemotherapyTHWeekly",
       "label": "Chemotherapy TH Weekly",
-      "action": [
-        {
+      "action": {
           "id": "1",
           "type": "create",
           "description": "Chemotherapy TH Weekly",
@@ -93,8 +92,7 @@
               "text": "Chemotherapy (procedure)"
             }
           }
-        }
-      ],
+        },
       "transitions": [
         {
           "id": "1",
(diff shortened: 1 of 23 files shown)


KeeyanGhoreshi:
Looks good to me overall.  I was testing it out and ran into an error that I could not reproduce, not sure if it's related to this PR or not, but it did have to do with `action` being undefined.  I'll try to see if I can recreate it.

src/types/pathways-model.d.ts
```diff
@@ -37,7 +37,7 @@ declare module 'pathways-model' {
   export interface ActionNode extends PathwayNode {
     cql: string;
     elm?: ElmLibrary;
-    action: Action[];
+    action: Action;
```
dehall, line 40: I get why we want this to be required, but since there are currently some examples of places where it's functionally null, just an empty object, it might make more sense just to have it be optional. The `if (action !== ({} as Action))` that I see in a few places is just a weirder looking null check.

blangley28: So this was a little bit more complex than just making it optional since we use `action` to determine if the node is an action node or not. Luckily `null !== undefined` in JS so I made it `Action | null`. There was a lot which needed to be updated with this but it looks like its all working well

blangley28:
Is anyone else having these two issues too:
1. When adding details to an action node it doesn't automatically open. I know Noranda fixed this but I can't seem to figure out how I broke it.
2. Do icons appear for you on nodes? It seems to be 50/50 where sometimes when I add a new node and select a type the icon never appears (but once I return to the pathway selection page then open the pathway again its there). It seems to happen randomly and I haven't been able to reproduce recently. Has anyone else experienced this too?

dehall:
I didn't try every possible combination but I'm not seeing any issues with node icons
.

Is this all set & ready for review?

public/static/pathways/her2_pathway.json
```diff
@@ -190,13 +184,13 @@
     "NoEndocrineTherapy": {
       "key": "NoEndocrineTherapy",
       "label": "No Endocrine Therapy",
-      "action": [],
+      "action": {},
```
dehall, line 187: This should be `null` now, not `{}` right? Same with the one a few lines below

dehall: Updated suggestion: get rid of the "No Endocrine Therapy" node altogether and put something meaningful in the "Endocrine Therapy" node, like a ServiceRequest for 169413002 | Hormone therapy (procedure)

dehall:
Overall looks good -- just the 2 nodes in Early Stage HER2+ that are actions without actions are causing trouble. Once those stop causing trouble, either with more engine changes ( 😞 ) or by making the nodes meaningful ( 🙂 ) I'll approve.

public/static/pathways/her2_pathway.json
blangley28: I'll populate them with something meaningful but I think this is important:
>With null action the builder thinks they are "observation" type.

They shouldn't be seen as "observation"...

blangley28: Made action required and updated the HER2+ pathway to have actions on all nodes (remove no endocrine therapy and add a service request for endocrine therapy)

dehall (approved):

KeeyanGhoreshi (approved):
looks good, works good

PR #46, opened 2020-09-09
CQL import and export

dehall:
This PR adds the ability to import CQL files (temporarily replacing the ability to import ELM files, until we figure out how to make that work) and includes the CQL in the exported CPG output.
Imported CQL is immediately sent to the CQL-to-ELM webservice. (no error handling yet) Any boolean-valued definitions are identified and stored as criteria. (If the ELM doesn't have return types, then it will import everything. Note also that the **entire** CQL is stored, because we don't know what other functions might be defined in there that the boolean functions reference)

CQL is exported as follows:
 - Criteria constructed in the builder are included in the "main" Library, which is the one linked in `PlanDefinition.library`
 - Criteria imported from CQL are exported as another Library, referenced in CQL in the "main" Library
 - Applicability conditions in CPG are always the name of a defined statement. We should mark these as "text/cql.name" but typescript doesn't like that so instead we should reference the library function directly, ex `Lib0123.DefinedThing` instead of just `DefinedThing`

So for instance, assume you have a CQL library "utils.cql" with one function: "IsMale".
1. Import this CQL library
2. Create a pathway and a branch on the pathway
3. Transition 1 on the branch = IsMale
4. Transition 2 on the branch = create a new Criteria for Gender: Female
5. Add an action to both nodes after the branch
6. Export the pathway as CPG

You should see 2 Libraries, containing base-64'ed CQL.

The main library will contain something along the lines of:

```
library LIBabcdefg version '1.0'

using FHIR version '4.0.0'

include "utils" version '1.0' called utils

context Patient

define "IsMale": utils.IsMale

define "female": Patient.gender.value = 'Female'
```

The second library will contain the exact contents of utils.cql.

Based on some experimentation, this approach will work in cqf-ruler. (reviewers please try it out, if possible. that may require messing with the PlanDefinitions so there are no nested PlanDefs until we get that working)


blangley28 (commented):
Took a look through - haven't tested yet. Only issue I see off the bat is the additional libraries. Do we need to make the main library point to them or will CQF Ruler automatically consume them all. CPG IG constrains library to be 0..1

dehall:
> Took a look through - haven't tested yet. Only issue I see off the bat is the additional libraries. Do we need to make the main library point to them or will CQF Ruler automatically consume them all. CPG IG constrains library to be 0..1

From my testing cqf-ruler seemed to import everything nested as needed. Starting with something that this branch can export, as of now the only thing that needs manual post-processing is the "text/cql.name" thing. But I'm looking into how to make that work.

dehall:
To reviewers: if you want to try out the exported CPG in cqf-ruler, A) make sure that all transitions actually lead somewhere and the target action has something on it like a medication or procedure request, and B) modify the exported top-level PlanDefinition so that each `action` points to an ActivityDefinition instead of to another PlanDefinition in the `definitionCanonical` field.

src/utils/cpg.ts
```diff
@@ -164,14 +185,93 @@ function createLibrary(pathway: Pathway): Library {
...
+  // for each one, add the criteria CQL to our appropriate map
+  //  - if it's an included library, keep track of the library name and version
+  //    as well as the specific definition we're referencing
+  //  - if it was constructed in the builder, track the name and raw CQL
+  for (const nodeId in pathway.nodes) {
+    const node = pathway.nodes[nodeId];
+    for (const transition of node.transitions) {
+      if (transition.condition?.criteriaSource) {
+        const criteriaSource = criteria.find(c => c.id === transition.condition?.criteriaSource);
+        if (criteriaSource && criteriaSource.elm && criteriaSource.cql) {
```
radamson, line 207:
```suggestion
        if (criteriaSource?.elm && criteriaSource?.cql) {
```

src/components/CriteriaProvider.tsx
```diff
@@ -40,16 +41,20 @@ const DEFAULT_ELM_STATEMENTS = [
...
-function builderModelToCriteria(criteria: BuilderModel): Criteria {
+function builderModelToCriteria(criteria: BuilderModel, label: string): Criteria {
   return {
     id: shortid.generate(),
-    label: `${criteria.type}${Date.now()}`,
+    label,
     modified: Date.now(),
     builder: criteria,
-    statement: criteria.cql
+    statement: label
```
radamson, line 50: Just curious, but is there a reason `label` and `statement` are the same now and do we ever expect them to be different?

radamson (commented):
Haven't tested yet, but code looks good.  Noticed a few small nits.

src/utils/cpg.ts
```diff
@@ -164,14 +185,93 @@ function createLibrary(pathway: Pathway): Library {
...
+          const libraryIdentifier = criteriaSource.elm.library.identifier;
+          includedCqlLibraries[libraryIdentifier.id] = {
+            cql: criteriaSource.cql,
+            version: libraryIdentifier.version
+          };
+
+          referencedDefines[transition.condition.cql] = libraryIdentifier.id;
+        } else if (criteriaSource && criteriaSource.builder) {
```
radamson, line 215:
```suggestion
        } else if (criteriaSource?.builder) {
```

src/components/CriteriaProvider.tsx
dehall: `label` is how it's displayed in the UI, and `statement` is the way it's referenced in CQL. So when we create a Criteria it's the same, but I think they will be different for imported criteria. (`label` in that case contains the name of the library that it was imported from) 
IIRC `label` is display-only, so changing the above is mostly a display fix for how the UI displays built criteria, which as of today in master looks a little weird: 
image

src/utils/cpg.ts
```diff
@@ -46,12 +47,23 @@ export function createActivityDefinition(action: Action): ActivityDefinition {
...
+      break;
+    case 'ServiceRequest':
+    case 'CarePlan':
+      delete activityDefinition.productCodeableConcept; // not allowed on these in cqf-ruler
+      activityDefinition.code = action.resource.code;
+      break;
+    default:
+      // do nothing
+      break;
+  }
```
blangley28, line 65: Is there a reason `productCodeableConcept` is left here. The `ActivityDefinition` type adds `productCodeableConcept` as a required field but since it isn't anymore the model can be updated then simplify this code here
```suggestion
  };

  switch (kind) {
    case 'MedicationRequest':
      activityDefinition.productCodeableConcept = action.resource.medicationCodeableConcept;
      break;
    case 'ServiceRequest':
    case 'CarePlan':
      activityDefinition.code = action.resource.code;
      break;
    default:
      // do nothing
      break;
  }
```

blangley28 (changes requested):
Since we now require cql the existing criteria needs to be updated to include its cql otherwise the exported CQL library is just
```
library LIBabce107 version '1.0'

using FHIR version '4.0.1'



context Patient
```

dehall:
So you're totally right about the existing preloaded ELM criteria -- this does kind of make them useless. But just putting a CQL file into the criteria folder won't work with the existing Service Implementation which expects everything to be JSON. I'll have to think about what the best way to make that work is. I'm open to suggestions

dehall:
Ok the last commit deleted a bunch of things that we aren't actually using anymore (most of the cql-to-elm and static CQL resources). The important thing is that it now loads the demo CQL from the `public/static/cql` folder as fixtures, instead of loading any ELM.

radamson:
This is probably beyond the scope of this PR, but how do we want to handle duplicate criteria?  I can upload the same criteria multiple times and see something like:

Screen Shot 2020-09-17 at 1 23 25 PM

which also appears in the criteria selection when authoring a pathway:

Screen Shot 2020-09-17 at 1 24 15 PM

This is an edge case, but something to consider as part of the design.  Not necessarily looking for an answer, but just wanted to bring it up because I'll forget otherwise...

dehall:
Yeah let's make a separate task for identifying duplicates. We could start with something simple like "nothing can have the same name" but there could easily be false positives with that so not a permanent solution.

radamson:
> Yeah let's make a separate task for identifying duplicates. We could start with something simple like "nothing can have the same name" but there could easily be false positives with that so not a permanent solution.

Created PATHWAYS-367 for this.

blangley28:
I'm assuming this isn't trivial so I can make a new task but wanted to bring it up now: 
When I upload the following CQL file the `TumorSize` definition is loaded
```
library "Tumor-Size-2cm-and-Under" version '1.0.0'
using FHIR version '4.0.0'
include "CDS_Connect_Commons_for_FHIRv400" version '1.0.1' called C3F 

valueset "Tumor Size VS": '2.16.840.1.113762.1.4.1116.554'

context Patient
define "Tumor Size":
  C3F.QuantityValue(C3F.MostRecent([Observation: "Tumor Size VS"])) <= 2 'cm'
```

But as soon as I add another statement only the `2cmAndUnder` statement is imported.
```
library "Tumor-Size-2cm" version '1.0.0'
using FHIR version '4.0.0'
include "CDS_Connect_Commons_for_FHIRv400" version '1.0.1' called C3F 

valueset "Tumor Size VS": '2.16.840.1.113762.1.4.1116.554'

context Patient
define "Over2cm":
  C3F.QuantityValue(C3F.MostRecent([Observation: "Tumor Size VS"])) <= 2 'cm'

define "2cmAndUnder":
  exists([Observation: "Tumor Size VS"])
```

If I update the file above to use 
```
define "Over2cm":
  exists([Observation: "Tumor Size VS"])
```
then both statements are imported.

I'm assuming the issue here is that no types are included in the ELM for the C3F version but as soon as the second statement is included the ELM has a type and its not boolean for the first statement since the C3F library was not included in the translation. So I _think_ the solution here is to require C3F to be imported with this library but that would require more checks on upload. Perhaps for now we just include the CDS connect libraries so when we convert to ELM they are automatically included?

dehall:
Yeah that really gets to the broader question of importing CQL that references another library. Without having that referenced C3F library I think you could make the argument that the first CQL library isn't even valid. (There must be some warnings or errors generated by the cql-to-elm webservice)
My preference would be to make that another task because I think it will need some thought to get everything right, in terms of UI and data model. (for example, if library A references library B, can you upload A then B? Or does it have to be B then A? Or do you have to upload them together in a single zip file or something? Is someone uploading CQL going to know how they depend on one another?)

blangley28:
Yes I agree completely. Created [PATHWAYS-369](https://jira.mitre.org/browse/PATHWAYS-369)

blangley28 (approved):
LGTM - can import correctly and tested against the cqf-ruler

radamson (approved):
Incredible.

PR #52, opened 2020-09-25
Edit Existing Criteria

danlee1025:
Added functionality to edit existing criteria
    - Criteria builder will automatically pre-populate fields after user clicks on edit button

image

### DIFF
--- a/src/components/CriteriaBuilderProvider.tsx
+++ b/src/components/CriteriaBuilderProvider.tsx
@@ -7,6 +7,7 @@ import React, {
   FC,
   useContext
 } from 'react';
+import { BuilderModel } from 'criteria-model';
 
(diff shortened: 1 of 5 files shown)


mdube7:
This maybe should be its own issue but if you build a criteria and then use that for a different transition the criteria name does not populate

mdube7 (commented):
When editing the criteria itself, i.e. values of age range, it will update the criteria for all locations this criteria is used.
When you edit the criteria name, it will update the criteria name in the `Select Criteria` dropdown but won't update the name in other locations where this criteria is used.

mdube7 (approved):
Looks good to me!

dehall (approved):

PR #53, opened 2020-10-02
add reference node (wip)

KeeyanGhoreshi:
Adds the reference node.  Very straightforward, adds another node type and then sets up the conditions for how to handle the new node type everywhere that the branch/action nodes are already handled.  I've also added a "type" property to all the nodes, so that they can be easily differentiated without needing complex logic.

The only thing that needs work is the cpg output.  I'm not sure exactly what's supposed to be output for the cpg, so for now this is a "wip".


src/components/Sidebar/Sidebar.tsx
```diff
@@ -109,7 +153,39 @@ const Sidebar: FC = () => {
                 value=""
               />
             )}
-
+            {nodeType === 'reference' && (
+              <div>
```
blangley28, line 157: To clean up this file the other node types have a an Editor component if its more than just a single line. I would suggest breaking this out into `ReferenceNodeEditor.tsx`

src/components/Sidebar/Sidebar.tsx
```diff
@@ -109,7 +153,39 @@ const Sidebar: FC = () => {
...
+                <div className={styles.referenceDropdown}>
+                  <DropDown
+                    id="reference"
+                    label="Pathway Reference"
+                    options={pathwayOptions}
+                    onChange={selectPathwayReference}
+                    value={(currentNode as ReferenceNode).referenceId}
+                  />
+                  <div className={styles.referenceChevron}>
+                    <Link
```
blangley28, line 174: I noticed the node redirects in the sidebar just use `history.push(url)` instead of this `Link` component. Unless there's a reason to use this over the redirect I think we should stay consistent. See `utils/nodeUtils.tsx redirect()` for how redirecting nodes works

src/types/pathways-model.d.ts
```diff
@@ -32,6 +32,7 @@ declare module 'pathways-model' {
     label: string;
     transitions: Transition[];
     nodeTypeIsUndefined?: boolean;
+    type: 'action' | 'branch' | 'reference' | 'other';
```
blangley28, line 35: What falls under the `other` category? I noticed null nodes still use `nodeTypeIsUndefined`. Perhaps we can consolidate these two together
```suggestion
    type: 'action' | 'branch' | 'reference' | 'null' | 'start';
```

blangley28 (changes requested):
Few minor comments but looks good and works well. We can discuss the CPG export but this is where nested plan definitions will come in (the set of all referenced pathways will make up a strategy while each individual pathway will use the current exporter to create a recommendation definition).
Two general comments
1) Maybe its just me but not having a line break after your methods makes it harder to read
2) If you load the test pathway and click on any of the nodes without types (ie. Radiation) you can't add a type in the sidebar. I think this is because it has type 'other' but does not have `nodeTypeIsUndefined`

src/types/pathways-model.d.ts
dehall: Looks like 'other' just means start nodes, so I agree with making that explicit. Did you mean `'null'` or just `null` here?

KeeyanGhoreshi:
still a bit shaky on what exactly we wanna see out of the CPG export.  The referenced pathways show up, but I'm not sure if they show up the right way.

blangley28:
> The referenced pathways show up, but I'm not sure if they show up the right way.

So it seems like all the ActivityDefinitions are being added in the correct place (top level resources in Bundle.entry). But I only see two PlanDefinitions. This is how you have have it organized:
```
Bundle.entry
|- StrategyDefinition
     |-action
          |-PlanDefinition
```
The StrategyDefinition is also named PathwayA while the RecommendationDefinition is named PathwayB. The correct structure is
```
Bundle.entry
|- StrategyDefinition
     |- action
          |- PlanDefinition (PathwayA)
          |- PlanDefinition (PathwayB)
```
Let me work on an example for you. I think that might help.

public/static/pathways/her2_pathway.json
```diff
@@ -26,7 +26,8 @@
            "id": "1",
            "transition": "NodeStatus"
          }
-       ]
+       ],
+       "type": "other"
```
dehall, line 30: This should probably be changed to `"start"` now, along with any remaining `"other"`s

KeeyanGhoreshi:
ok so now the cpg export should be putting things in the right place.

Though there are still some outlying questions.  Like what happens when Pathway A conditionally references Pathway B an C, but pathway C also references pathway B conditionally?  The conditions aren't tied to a specific pathway reference.

blangley28:
> What happens when Pathway A conditionally references Pathway B and C, but pathway C also references pathway B conditionally?

So ideally I think the result should be this:
1) Check if Pathway B has already been exported to a RecommendationDefinition
2) If it has don't convert it again (the resource is already in the bundle), but still add the reference in StrategyDefinition.action

This will probably require another data structure (probably just a map from PathwayId to RecommendationId)

blangley28:
I can't take a picture of it but the arrow in the sidebar next to pathway reference (which brings you to that pathway) should make the cursor a pointer like the other arrow buttons

blangley28:
Everything looks good to me in the exported CPG file except for two small details:
1. The StrategyDefinition.action don't reference the RecommendationDefinitions (via `definitionCanonical`)
2. The Library for PathwayB only has CQL and not ELM (or preconditions). The Library for PathwayA has both and when I export only PathwayB it too has both.

Exported Files using the new CPG exporter: [CPG Pathways.zip](https://github.com/DBCG/cql_engine/files/5355263/CPG.Pathways.zip)
 

For everyone else's reference here are the documents I gave Keeyan:
Screen Shot 2020-10-07 at 9 24 22 AM
Screen Shot 2020-10-07 at 9 24 44 AM

src/utils/nodeUtils.ts
```diff
@@ -8,11 +17,13 @@ export function isActionNode(node: PathwayNode): node is ActionNode {
   return action !== undefined;
```
blangley28, line 17: Suggest updating this to also use comparison to `node.type`

src/utils/builder.ts
```diff
@@ -375,13 +399,13 @@ export function setActionNodeElm(pathway: Pathway, nodeKey: string, elm: ElmLibr
   });
 }
 
-export function getNodeType(node?: ActionNode | BranchNode | PathwayNode | null): string {
-  if (!node || node.nodeTypeIsUndefined) {
+export function getNodeType(
```
blangley28, line 402: I know this method was here before your PR but this method is better suited for `utils/nodeUtils` than this file

src/utils/cpg.ts
```diff
@@ -74,64 +75,107 @@ export class CPGExporter {
...
-      this.pathway.description,
-      'recommendation',
-      library.id
+      `Strategy${this.pathway.name}`,
+      `Strategy for ${this.pathway.name}`,
+      'strategy'
     );
 
-    /*
+    this._export(strategyDefinition, this.pathway);
```
blangley28, line 89: `strategyDefinition` could probably be a class attribute if you want so it doesn't need to be passed here. I'm also fine with the way it is now if you prefer

src/utils/cpg.ts
```diff
@@ -74,64 +75,107 @@ export class CPGExporter {
...
+  _export(
+    strategyDefinition: PlanDefinition,
+    pathway: Pathway,
+    referencedFrom?: PlanDefinitionAction,
+    condition?: PlanDefinitionCondition[]
+  ): void {
+    // the action for the strategy definition
+    const pathwayAction = createAction(pathway.name, `Action for ${pathway.name}`);
+
+    if (referencedFrom && referencedFrom.id && pathwayAction.id) {
```
blangley28, line 103: This can be consolidated. Also I think `pathwayAction.id` will always be set right? Is this just an issue with the typing being optional?
```suggestion
    if (referencedFrom?.id && pathwayAction.id) {
```

src/utils/cpg.ts
KeeyanGhoreshi: im partial to the way it is now, I think it makes the functions more modular

src/utils/cpg.ts
KeeyanGhoreshi: yes id is optional

blangley28: Should it be? Is there ever a point in the builder where it isn't set? The type can be updated in `fhir-objects.d.ts`

dehall (commented):
Code looks good overall. 
When testing with the "Test Pathway (Example)" some of the nodes have no type and you can't edit them -- that needs to be fixed. Not sure if that's an issue with the pathway content or the code.
image

dehall (approved):
LGTM. As Ben mentioned the cursor should be a pointer on the arrow to go to the referenced pathway, but I'll approve now on the assumption that will get fixed tonight

PR #54, opened 2020-10-06
Import Pathway

radamson:
Adds the ability to import a pathway from the UI.  Currently there are no checks for ensuring that the uploaded pathway is actually valid.

Screen Shot 2020-10-06 at 9 24 47 AM


src/components/FileImportModal/FileImportModal.tsx
```diff
@@ -1,48 +1,51 @@
...
+    [onClose, onSelectFile]
   );
 
   const handleChooseFile = useCallback(() => {
     if (importFileRef?.current?.files?.[0]) setFileName(importFileRef.current.files[0].name);
   }, [importFileRef]);
 
   return (
-    <Dialog open={open} onClose={onClose} aria-labelledby="import-criteria" fullWidth maxWidth="sm">
+    <Dialog open={open} onClose={onClose} aira-labelledby="import-pathway" fullWidth maxWidth="sm">
```
dehall, line 48: Somehow a couple typos got introduced here.... `aria` -> `aira` here and `types` -> `tpyes` a few lines below

src/components/PathwaysProvider.tsx
```diff
@@ -43,6 +44,23 @@ export const PathwaysProvider: FC<PathwaysProviderProps> = memo(function Pathway
     [setPathways]
   );
 
+  const addPathwayFromFile = useCallback(
+    (file: File) => {
+      const reader = new FileReader();
+      reader.onload = (event: ProgressEvent<FileReader>): void => {
+        if (event.target?.result) {
```
dehall, line 51: Duplicated `if` here. Sometimes you just really want to make extra sure it's not null

src/components/FileImportModal/FileImportModal.tsx
radamson: oops

src/components/PathwaysProvider.tsx
radamson: just wanted to... double check

src/components/CriteriaList/CriteriaList.tsx
```diff
@@ -6,14 +6,17 @@ import { Button } from '@material-ui/core';
 import { usePathwaysContext } from 'components/PathwaysProvider';
 import Loading from 'components/elements/Loading';
 import CriteriaTable from './CriteriaTable';
-import ImportCriteriaModal from './ImportCriteriaModal';
 
 import useStyles from './styles';
+import FileImportModal from '../FileImportModal';
+import { useCriteriaContext } from '../CriteriaProvider';
```
blangley28, line 12: Should these be absolute imports?

src/components/PathwaysList/PathwaysList.tsx
```diff
@@ -23,19 +27,46 @@ const PathwaysList: FC = () => {
...
-      >
-        Create Pathway
-      </Button>
+      <div className={styles.buttonRow}>
+        <Button
+          className={styles.createPathwayButton}
+          variant="contained"
+          color="primary"
+          startIcon={<FontAwesomeIcon icon={faFileImport} />}
+          onClick={(): void => setImportPathwayOpen(true)}
```
blangley28, line 45: Use a callback here instead of passing in an anonymous function (https://mitrecorp.slack.com/archives/CLH34KL6A/p1594649905003300)

src/components/PathwaysList/PathwaysList.tsx
```diff
@@ -23,19 +27,46 @@ const PathwaysList: FC = () => {
...
+          startIcon={<FontAwesomeIcon icon={faPlus} />}
+          onClick={openNewPathwayModal}
+        >
+          Create Pathway
+        </Button>
+      </div>
+
+      <FileImportModal
+        open={importPathwayOpen}
+        onClose={(): void => setImportPathwayOpen(false)}
```
blangley28, line 63: Use callback here too

blangley28 (commented):
I know you're still working on the criteria stuff but I just wanted to make sure you were aware of this - when I import the pathway the criteria dropdown is not selected. The bottom criteria (Male) was built. In this situation are you planing on importing it as a new criteria?
Screen Shot 2020-10-08 at 12 45 05 PM

src/components/CriteriaList/CriteriaList.tsx
radamson: They..._absolutely_ should be.

src/utils/cpg.ts
```diff
@@ -77,6 +77,23 @@ export class CPGExporter {
     this.libraries = this.createLibraries();
   }
 
+  modelExport(): Pathway {
```
blangley28, line 80: Two comments about this:
1) Everything else in this file is specific to CPG. It seems a little weird to create a new `CPGExporter` instance and then export to something other than CPG.
2) There was already the `exportPathway` method in `utils/builder.tsx` which can be modified

src/types/pathways-model.d.ts
```diff
@@ -8,7 +8,7 @@ declare module 'pathways-model' {
     id: string;
     name: string;
     description: string;
-    library: string;
+    library: string | string[];
```
blangley28, line 11: Is there a reason this was made to be an array? I can't find anywhere it is being used as an array and I'm worried it may have some negative effects on the CPG exporter which only allows a single library to be referenced here

blangley28 (commented):
Didn't realize you were still working on the task so I reviewed again - I think these two comments are still worth noting for now though (just in case you were planning on building off them later)

src/types/pathways-model.d.ts
radamson: Turned this into an array so that all the CQL Libraries can be exported too.  If a pathway uses the "TNM" and "Tumor_Markers" Criteria Libraries then they'll be exported in with the pathway in this array.

Do you have a suggestion for how to include them otherwise?  It seems like the model (at least the exported representation) will need to change somehow.

src/utils/cpg.ts
radamson: I was just working in this class for now because I was using some of its features (e.g. it gathers up all the criteria libraries).  Ultimately I think this class would have become more general or it would have moved.  I'll take a look at the version in `utils/builder.tsx` though...didn't realize that was there 👀 .

radamson: I'll keep this in mind though and rework it.

dehall:
Discussed with @radamson this morning, and we decided on 4 tasks we still need to address:

 - [ ] import any referenced libraries into the CriteriaProvider
 - [ ] re-assign transition CQL/criteriaSource after importing the referenced libraries
 - [ ] the "main" library should only include CQL that was constructed with the criteria builder (anything else should be referenced directly in its original library from the `transition.condition..cql` field )
 - [ ] clean up the cpg.ts logic and make it available to both the CPG exporter and Camino exporter

src/utils/cpg.ts
dehall: I updated this so there's a new exporter class for the Camino model. Eventually I think we should extract the logic into 3 classes - CPGExporter, CaminoExporter, and common, but I'd rather wait until after Keeyan's PR is merged to avoid a potentially painful rebase.

src/types/pathways-model.d.ts
dehall: Rather than either type, I'd suggest we just make it an array. IMO it's easier to deal with an array of 1 entry rather than check if it's an array or string everywhere.

radamson: Made library just an array in 6d7ea8b.

src/utils/builder.ts
```diff
@@ -66,6 +67,9 @@ export function exportPathway(
   if (cpg) {
     const exporter = new CPGExporter(pathwayWithElm, pathways, criteria);
     pathwayToExport = exporter.export();
+  } else {
+    const exporter = new CaminoExporter(pathwayWithElm, criteria);
```
blangley28, line 71: Do we want the CaminoExporter to consider pathway reference nodes? I don't think that should be part of this task but perhaps a back log item to download all of them in a zip file

blangley28 (commented):
Looking really good. Was able to create a pathway, export it, and then re-import it again. Two general additional comments:
1. The cql libraries are added as new criteria. If its not too difficult perhaps a checksum can be done on the contents and if its the same as an existing criteria it links to that instead.
2. For the 3 pathways in public/static/pathways the library property needs to be updated to be a list instead of a string

src/utils/CaminoExporter.ts
```diff
@@ -0,0 +1,123 @@
...
+    cql: string;
+    version: string;
+  };
+}
+
+export class CaminoExporter {
+  pathway: Pathway;
+  criteria: Criteria[];
+
+  // TODO: figure out if elm works correctly here
```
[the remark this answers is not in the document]
dehall: I think this was just a copy & paste from the CPGExporter that I forgot to clean up, will delete unless @radamson gets there first

src/utils/builder.ts
dehall: Yeah let's make that a separate task. I was talking to Reece about that on slack:
> The pathway -> cpg translation creates new IDs at export time, so if you exported A then B to CPG separately, any reference from A to B would be lost, so we have to export them together. Whereas with the internal data model it only links by name, so if you export A then B, the relationship is preserved.

blangley28: Created [PATHWAYS-381](https://jira.mitre.org/browse/PATHWAYS-381)

src/utils/CaminoExporter.ts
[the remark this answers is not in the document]
radamson: deleted the TODO

src/utils/CaminoExporter.ts
```diff
@@ -0,0 +1,81 @@
...
+            const libraryIdentifier = criteriaSource.elm.library.identifier;
+            includedCqlLibraries[libraryIdentifier.id] = {
+              cql: criteriaSource.cql,
+              version: libraryIdentifier.version
+            };
+
+            referencedDefines[transition.condition.cql] = libraryIdentifier.id;
+
+            // prepend the library name if not already done
+            if (!transition.condition.cql.startsWith(libraryIdentifier.id)) {
```
dehall, line 50: Technically I think there's a chance of false positives here, but more importantly I would have expected the importer to remove the library id for consistency. Nothing else breaks when it's present after importing?

public/static/pathways/her2_pathway.json
```diff
@@ -2,7 +2,7 @@
    "id": "tB4jsxTDL",
    "name": "Early Stage HER2+ (Draft)",
    "description": "Pathway for the Early Stage HER2+ Initial Medical Consult Breast Cancer Pathway. This is a draft pathway.",
-   "library": "",
+   "library": [""],
```
blangley28, line 5: Should this just be an empty list instead?
```suggestion
   "library": [],
```
Same goes for neoadjuvant_pathway.json

src/utils/CaminoExporter.ts
radamson: I didn't see anything else break because most of the other functionality is keyed off `criteriaSource` as far as I know.  I'll make the change you suggested.

public/static/pathways/her2_pathway.json
radamson: Good call, changed.

src/utils/CaminoExporter.ts
dehall: After talking with Reece, it seems strange that nothing breaks, and apparently it's a lot easier to do this than to remove it as part of the import process, so in the interest of getting this in soon, we'll leave this as is

radamson: So I started working on changing this, but its a little trickier than I expected.

Right now re-associating the CQL requires that the library be prefixed onto the CQL so that it can find the criteria that matches the library/statement combo.  All the existing sample pathways are organized this way too.  I think we'd need to modify the data model to store the information linking the CQL to the library otherwise.  This information has to left on the pathway and I can't have the importer remove it because the re-association is done in a `useEffect` hook after the fact because setting the new criteria is asynchronous.

mdube7 (approved):
Looks good to me

blangley28 (commented):
One last change - if the same pathway is uploaded again (ie same id) it should not upload the pathway and alert the user (I'm fine with a JS alert) that the pathway has not been imported because it already exists.

blangley28 (approved):

PR #55, opened 2020-10-13
Undo

blangley28:
Implement undo and redo for pathway changes.

Modified the app to use the CurrentPathwayProvider for changes. This provider then synchronizes with the PathwayProvider (list of all pathways). All of the functionality has remained the same with the one exception that if the nodeid in the url does not exist the url does not update to 'Start'. The start node is still active but the url shows the nodeid which does not exist.

### DIFF
--- a/src/components/BuilderRoute.tsx
+++ b/src/components/BuilderRoute.tsx
@@ -9,7 +9,7 @@ import { useCurrentNodeContext } from './CurrentNodeProvider';
 const BuilderRoute: FC = () => {
   const { id, nodeId } = useParams();
   const { pathways } = usePathwaysContext();
-  const { setPathway } = useCurrentPathwayContext();
+  const { pathway: currentPathway, setPathway, resetPathway } = useCurrentPathwayContext();
   const { setCurrentNode } = useCurrentNodeContext();
   const pathwayId = decodeURIComponent(id);
   const pathwayIndex = useMemo(() => pathways.findIndex(pathway => pathway.id === pathwayId), [
@@ -20,8 +20,9 @@ const BuilderRoute: FC = () => {
   const currentNode = pathway?.nodes?.[decodeURIComponent(nodeId)];
 
   useEffect(() => {
-    setPathway(pathway);
-  }, [pathway, setPathway]);
+    if (currentPathway?.id === pathway?.id) setPathway(pathway);
+    else resetPathway(pathway);
+  }, [pathway, setPathway, resetPathway]);
 
   useEffect(() => {
     setCurrentNode(currentNode);
--- a/src/components/BuilderRoute.tsx
+++ b/src/components/BuilderRoute.tsx
@@ -9,7 +9,7 @@ import { useCurrentNodeContext } from './CurrentNodeProvider';
 const BuilderRoute: FC = () => {
   const { id, nodeId } = useParams();
   const { pathways } = usePathwaysContext();
-  const { pathwayRef, setPathway, resetPathway } = useCurrentPathwayContext();
+  const { pathwayRef, setCurrentPathway, resetCurrentPathway } = useCurrentPathwayContext();
   const { setCurrentNode } = useCurrentNodeContext();
   const pathwayId = decodeURIComponent(id);
   const pathwayIndex = useMemo(() => pathways.findIndex(pathway => pathway.id === pathwayId), [
@@ -20,9 +20,9 @@ const BuilderRoute: FC = () => {
   const currentNodeId = decodeURIComponent(nodeId);
 
   useEffect(() => {
-    if (pathway && pathwayRef?.current?.id === pathway.id) setPathway(pathway);
-    else if (pathway) resetPathway(pathway);
-  }, [pathway, pathwayRef, setCurrentNode, setPathway, resetPathway]); // eslint-disable-line
+    if (pathway && pathwayRef?.current?.id === pathway.id) setCurrentPathway(pathway);
+    else if (pathway) resetCurrentPathway(pathway);
+  }, [pathway, pathwayRef, setCurrentNode, setCurrentPathway, resetCurrentPathway]);
 
   useEffect(() => {
     if (pathwayRef.current?.nodes[currentNodeId])
--- a/src/components/BuilderRoute.tsx
+++ b/src/components/BuilderRoute.tsx
@@ -9,12 +9,7 @@ import { useCurrentNodeContext } from './CurrentNodeProvider';
 const BuilderRoute: FC = () => {
   const { id, nodeId } = useParams();
   const { pathways } = usePathwaysContext();
-  const {
-    pathway: currentPathway,
-    pathwayRef,
-    setPathway,
-    resetPathway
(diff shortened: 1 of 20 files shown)


KeeyanGhoreshi:
seems to work well. There's a small bug where if you create a node, then undo it's creation, then redo its creation, you can't open the node in the sidebar until you click on a different node.  Probably has to do with the URL not updating?  The pathway brings the START node into focus after undoing but the URL doesn't change, so I figure that is a problem.

src/components/Navigation/Navigation.tsx
```diff
@@ -38,10 +39,18 @@ const Navigation: FC = () => {
 
         <span className={styles.pathwayName}>{pathway?.name}</span>
       </div>
-      <IconButton onClick={openMenu} aria-controls="pathway-options-menu" aria-haspopup="true">
-        <FontAwesomeIcon icon={faEllipsisH} className={styles.navigationIcons} />
-      </IconButton>
-      <ExportMenu anchorEl={anchorEl} closeMenu={closeMenu} />
+      <div>
+        <IconButton onClick={undoPathway} disabled={!canUndoPathway} aria-label="undo">
```
KeeyanGhoreshi, line 43: it would be immensely useful to have ctrl-z/y shortcuts to do this as well

src/components/BuilderRoute.tsx
```diff
@@ -9,26 +9,30 @@ import { useCurrentNodeContext } from './CurrentNodeProvider';
 const BuilderRoute: FC = () => {
   const { id, nodeId } = useParams();
   const { pathways } = usePathwaysContext();
-  const { setPathway } = useCurrentPathwayContext();
+  const { pathwayRef, setPathway, resetPathway } = useCurrentPathwayContext();
```
radamson, line 12: Looks like it was already this way, but it might be less confusing if `setPathway` was `setCurrentPathway`.  I know some of this will likely change when the backend is implemented so maybe it's not worth it.

What do you think?

src/components/BuilderRoute.tsx
```diff
@@ -9,26 +9,30 @@ import { useCurrentNodeContext } from './CurrentNodeProvider';
...
   const pathway = pathways[pathwayIndex];
-  const currentNode = pathway?.nodes?.[decodeURIComponent(nodeId)];
+  const currentNodeId = decodeURIComponent(nodeId);
 
   useEffect(() => {
-    setPathway(pathway);
-  }, [pathway, setPathway]);
+    if (pathway && pathwayRef?.current?.id === pathway.id) setPathway(pathway);
+    else if (pathway) resetPathway(pathway);
+  }, [pathway, pathwayRef, setCurrentNode, setPathway, resetPathway]); // eslint-disable-line
```
radamson, line 25: git won't let me insert a suggestion but I was able to lint fine without the magic `// esline-disable-line` comment.  Are you getting an error without it or can we remove this?

src/components/Sidebar/SidebarHeader.tsx
```diff
@@ -156,11 +154,11 @@ const SidebarHeader: FC<SidebarHeaderProps> = ({ node, isTransition = false }) =
               className={clsx(
                 styles.headerLabel,
                 styles.headerLabelText,
-                node.type === 'start' && styles.headerLabelStart
+                node.key === 'Start' && styles.headerLabelStart
```
radamson, line 157: why check against the `key` instead of the `type` like it was before?  `type` seems safer since its type is an enumerated list of strings instead of just a string.

radamson:
Dang this is cool.

I can't tell if its a little more laggy filling out descriptions, but it seems like a little.  Not bad at all though and with a production build I doubt it would be noticeable.  Going back with setting the description goes letter by letter which is kind of annoying, but not sure there is anything else we could easily do about that.

src/components/BuilderRoute.tsx
blangley28: Yea I was contemplating changing it due to my own laziness didn't - I shall go an update now though!

src/components/BuilderRoute.tsx
blangley28: This should be fine now. It was failing before but I fixed the error and forgot to remove the comment

src/components/Sidebar/SidebarHeader.tsx
blangley28: I think this was a rebase issue. Something else funky happened with these which I caught but I missed this one!

blangley28:
> I can't tell if its a little more laggy filling out descriptions, but it seems like a little. Not bad at all though and with a production build I doubt it would be noticeable. Going back with setting the description goes letter by letter which is kind of annoying, but not sure there is anything else we could easily do about that.

This is probably related to [PATHWAYS-387](https://jira.mitre.org/browse/PATHWAYS-387) where because there is no confirm button the state is being updated on each character. 

> I don't think the hot key's are technically right for mac, but they work with control! Not sure it's worth changing because I feel like different browsers/website might treat it differently anyways.

I added command+z and command+y as well. Just a note it doesn't work with the OS button in the top menu bar.

radamson (approved):
This is awesome.

Tested it out again and it worked well.  I like that if I go back to the pathways list and then come back to the pathway it still has the history to undo/redo.  If you click on another pathway while in the list it is understandably lost, but saving the history to enable that might be a cool thing in the future if users wanted it.  But that's overkill I'm sure.

noranda:
This looks good and works well! I have just a few UI suggestions:

1. Can we add a hover tooltip to the buttons (while it seems obvious what they do to us, we should be explicit for all users)?
2. Can we disable a button when it can't be used ?

Otherwise, nice job!

blangley28:
> 1. Can we add a hover tooltip to the buttons (while it seems obvious what they do to us, we should be explicit for all users)?

Yes I can add that!

> 2. Can we disable a button when it can't be used?

It is disabled - the material UI for IconButtons apparently doesn't change the styling when it is disabled (but the button does not work when it is disabled). I can add "(disabled)" to the tooltip

radamson:
There are some _**very**_ slight styling differences when it's disabled in that the background emphasis is gone.  It didn't really register with me when I was actively using it the same way a more typical greyed out button does though  ¯\_(ツ)_/¯

disabledbutton

blangley28:
Hopefully the tooltip makes it clear now whether or not the button is disabled

noranda:
> It is disabled - the material UI for IconButtons apparently doesn't change the styling when it is disabled (but the button does not work when it is disabled). I can add "(disabled)" to the tooltip

Ah yes I see that now! So the issue is the disabled styles are getting overwritten. To fix, remove `navigationIcons: { ... color: theme.palette.common.gray, ... }` from `src/components/Navigation/styles.txt` and then update the `src/styles/theme.tsx` file on line 199 to:

```
  MuiIconButton: {
    root: {
      color: colors.gray,
      '&:disabled': {
        color: colors.grayMedium
      }
    }
  }
```

noranda:
Looks great! Nice job!

noranda (approved):

PR #56, opened 2020-10-15
Selection Checkboxes for Pathways List

radamson:
Adds selection checkboxes for the pathways in the pathways list.  Pathways can then be deleted and exported as a group.

Potentially obnoxious gifs of it in action below:

**Selection:**

selection

**Export:**

export

**Delete:**

delete


noranda (changes requested):
Looking good! The select all button works well. Just a few issues:

1. The styling is a bit off, specifically the ripple circle is more oval. To fix:
Screen Shot 2020-10-19 at 11 16 38 AM

2. The action buttons (export, move to, delete) should only render if one or more checkboxes are selected, otherwise there are no actions to do (see design). Currently, if you try to export with nothing selected, you will export an empty zip. 

3. We should also include checkboxes in the Criteria List. Is this a separate task, or is it easy enough to add here?

src/components/PathwaysList/PathwaysListCheckbox.tsx
```diff
@@ -0,0 +1,62 @@
+import { ChangeEvent, useCallback, useMemo, useState } from 'react';
```
noranda, line 1: Since this is a hook and not a component, and since we will want to reuse it for the criteria list, why don't we create a `src/hooks` folder to keep it in. We could also move the `useRefState` hook from utils into this new folder as well.

src/components/PathwaysList/PathwaysList.tsx
```diff
@@ -38,28 +59,89 @@ const PathwaysList: FC = () => {
...
+      deletePathway(id);
+      setSelected(new Set());
+    });
+  }, [deletePathway, selected, setSelected]);
+
+  const [exporting, setExporting] = useState(false);
+
+  const handleExportAll = useCallback(() => {
+    setExporting(true);
+    const zip = new JSZip();
```
blangley28, line 73: If there's only one file selected do we want to download as just a json file? The logic can then be expanded for referenced pathways so pathways with references are exported as zips but standalone pathways are json

blangley28 (commented):
Tested all the functionality out and it works. Good to go after Noranda's comments are incorporated

src/components/PathwaysList/PathwaysList.tsx
noranda: My vote is yes on this.

noranda:
Looks great! One last comment/question. We should probably have the confirmation popup for when we delete pathways, like we do when you click on the delete button in the table. It might say "Are you sure that you would like to delete the selected pathway(s)?" If it's easy enough to add to this PR?

radamson:
I think all the comments up to this point have now been addressed.  As I was working through this I noticed that the CriteriaList and PathwayList components share a lot of the same code.  I started working on de-duplicating some of this code into some shared components, but it wasn't easy enough for me to finish just yet.  I think any code that refactors these components would be better suited as a separate PR for ease of review so I'll hold off on commit anything for that unless we decide we'd rather tackle that in this PR.

src/utils/builder.ts
```diff
@@ -35,22 +36,42 @@ export function createNewPathway(name: string, description: string, pathwayId?:
...
-): void {
-  const pathwayString = exportPathway(pathway, pathways, criteria, cpg);
-  // Create blob from pathwayString to save to file system
-  const pathwayBlob = new Blob([pathwayString], {
-    type: 'application/json'
-  });
+): Promise<void> {
+  // Determine if a single pathway or multiple pathways are being exported
+  const paths =
+    Array.isArray(pathway) && pathway.length > 1 ? pathway : ([] as Pathway[]).concat(pathway)[0];
```
blangley28, line 46: Is there a reason you use `([] as Pathway[]).concat(pathway)[0]` instead of just `pathway`. Seems like a very long way to get back what you started with....

It might also just be easier to make `pathway` always be a list and then just check the length in the if statement below.

blangley28 (commented):
Looks good, I like the new hook directory. I agree those changes should be put up in a separate PR just so its easier to review.

src/utils/builder.ts
radamson: `([] as Pathway[]).concat(pathway)[0]` is used in the case where only a single pathway and wraps that pathway in an array.  I'd be fine with making pathway always be a list, but other existing code already expects to be able to pass in a single pathway and might need similar checks anyways.

noranda (approved):

blangley28 (approved):

PR #57, opened 2020-10-16
Fix "read only property `criteriaSource`" Error

radamson:
My last PR broke a React cardinal rule...[Do Not Modify State Directly!](https://reactjs.org/docs/state-and-lifecycle.html#do-not-modify-state-directly). This PR attempts to right those wrongs.

Previously if one tried to update a criteria description without having a criteria selected this error would be seen: `TypeError: Cannot assign to read only property 'criteriaSource' of object '#<Object>'`


noranda (approved):
This looks good! Thanks for making the fix so quickly!

blangley28 (approved):
LGTM

PR #58, opened 2020-10-16
Transition label updates

noranda:
This PR satisfies [PATHWAYS-385](https://jira.mitre.org/browse/PATHWAYS-385) and does the following:

- Reduces the font size of the transition labels
- Increases the truncation size cutoff for transition labels from 12 characters to 20 characters
- Adds the ability to split transition labels into up to 3 lines using a comma `,` (adds helper text to input to indicate)

Before:
Screen Shot 2020-10-16 at 2 09 24 PM

After:
Screen Shot 2020-10-16 at 2 08 49 PM

To test: Import the attached pathway. Try editing any of the transition labels.

Note: This is branched off of [PR-57](https://github.com/mcode/pathway-builder/pull/57) which fixes a bug that was needed to build up a pathway used for testing. Therefore, this should be merged after that PR.

Note: There exists an issue with transition label placement (overlapping, crowding) when pathways become large. You will see this in the example pathway. I will add a separate JIRA task to deal with this issue as it was out of scope.

[NCCN Invasive Breast Cancer v5.2020 - BINV-15.json.zip](https://github.com/mcode/pathway-builder/files/5393517/NCCN.Invasive.Breast.Cancer.v5.2020.-.BINV-15.json.zip)

### DIFF
--- a/src/components/DagreGraph/GraphEdges.styles.ts
+++ b/src/components/DagreGraph/GraphEdges.styles.ts
@@ -16,7 +16,7 @@ export default makeStyles(
       height: '100%',
       zIndex: 1,
       '& text': {
-        fontSize: '18px'
+        fontSize: '14px'
       },
       '& #arrowhead-inactive polygon': {
         fill: theme.palette.common.blue
--- a/src/components/DagreGraph/GraphEdges.tsx
+++ b/src/components/DagreGraph/GraphEdges.tsx
@@ -13,9 +13,12 @@ interface EdgeProps {
 }
 
 interface LabelTooltipProps {
-  label: Label;
+  labelText?: string | null;
+  isEnabled: boolean;
   children: ReactElement;
 }
+
+const TRUNCATION_SIZE = 20;
 
 /**
  * The points to use in the Cubic command is determined by the length of points array
@@ -35,9 +38,9 @@ const generatePathString = (points: Coordinate[]): string =>
     `M ${points[0].x} ${points[0].y} `
   );
 
-const LabelTooltip: FC<LabelTooltipProps> = ({ label, children }) => {
-  if (label.text.length > 12) {
-    return <Tooltip title={label.text} children={children} />;
+const LabelTooltip: FC<LabelTooltipProps> = ({ labelText, isEnabled, children }) => {
+  if (isEnabled && labelText) {
+    return <Tooltip title={labelText} children={children} />;
   } else {
     return <>{children}</>;
   }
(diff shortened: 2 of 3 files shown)


radamson:
Tested it out and it worked well.  Had a couple thoughts while using it

* Is there any concern with having commas cause a new line on the display?  If a user actually wanted or needed to use a comma in the display they couldn't now.

* Is there any reason to not make `Criteria Display` an actual multi-line text area?  That way they could actually put in new line characters and have greater control over what is entered.

Thoughts on this?

noranda:
@radamson Thanks for reviewing!

1. I gave it some thought regarding which character to use and came to the conclusion that commas were optimal as they are not likely to be used in a label. I also considered using `-` or `/` but those seemed more likely to be used. We could offer an escape option if needed, but it's probably more work than it's worth for this task.

2. The reason I didn't use a text area field here is because of how the text is rendered. Since it's in an `svg`, it has to be parsed and put into `<tspan>` tags for each line. I don't know of a way to parse a string from a text area field to include line breaks. There is probably a way to do this, but again I was going for the easiest approach here to keep it in scope. If you can come up with an easy way to do this, let me know!

radamson (approved):
Makes sense.  This is a clear improvement so I think its worth getting this in and we can continue to explore using a text area, if Jim or someone else actually wants it, later.

blangley28 (commented):
I personally think this is harder to see than before.

The tag on the lower left is missing the "yp" because "HER2-targeted Therapy" overlaps.
Screen Shot 2020-10-19 at 9 28 52 AM

When that node is expanded it actually covers the entire label
Screen Shot 2020-10-19 at 9 30 35 AM

As you can see above there's also some very peculiar placement of labels. For example in the image below it looks like the right most label is for both of those crossing arrows when in reality it is only for "Ado-trastuzumab Emtansine" and the label all to the left "ypT0N0 or pCR" is for "HER2-targeted Therapy".
Screen Shot 2020-10-19 at 9 31 21 AM

I know its not perfect before this PR but I think it is easier to see which label goes to which transition:
Screen Shot 2020-10-19 at 9 34 49 AM

noranda:
@blangley28 This is definitely an issue and I created a new [JIRA task](https://jira.mitre.org/browse/PATHWAYS-391) to deal with this, however this is not related to the changes made in this PR, which doesn't change any of the positions of the labels. All this does is add up to 2 additional lines under the current position, and makes the text smaller and longer per line. You can see it better with this before and after:

Before:
Screen Shot 2020-10-19 at 11 54 47 AM

After:
Screen Shot 2020-10-19 at 11 54 11 AM

There are definitely issues with overlapping and crowding that make the pathway labels difficult to understand, so that needs to be addressed, but it's out of scope for this task.

blangley28 (approved):
Good to go with the creation of the new task. Looks like it needs a rebase to merge though.

PR #59, opened 2020-10-19
add autocomplete

KeeyanGhoreshi:
This uses the the material UI Autocomplete component to make our dropdowns searchable.  The only thing to note is that the pattern where dropdowns would have helper functions to extract the `events.target.value` has been gotten rid of.  I thought it was a bad pattern to begin with, but the autocomplete component also doesn't return an event with a value, it returns the value itself, so the `onChange` function needed to change to accept just the raw value string.  This just means that the helper functions in other components were removed.

You can test that everything works by typing in a value and selecting from the list.

### DIFF
--- a/src/components/Sidebar/BranchTransition.tsx
+++ b/src/components/Sidebar/BranchTransition.tsx
@@ -223,7 +223,7 @@ const BranchTransition: FC<BranchTransitionProps> = ({ transition }) => {
 
             <TextField
               label="Criteria Display"
-              style={{ width: '100%' }}
+              className={styles.criteriaText}
               value={transition.condition?.description || ''}
               variant="outlined"
               onChange={setCriteriaDisplay}
--- a/src/components/Sidebar/BranchTransition.tsx
+++ b/src/components/Sidebar/BranchTransition.tsx
@@ -84,10 +84,9 @@ const BranchTransition: FC<BranchTransitionProps> = ({ transition }) => {
   ]);
 
   const selectCriteriaSource = useCallback(
-    (event: ChangeEvent<{ value: string }>): void => {
+    (criteriaId: string): void => {
       if (!currentNodeRef.current || !pathwayRef.current) return;
 
-      const criteriaId = event?.target.value || '';
       const selectedCriteria = criteria.find(c => c.id === criteriaId);
       if (!selectedCriteria) return;
       const newPathway = setTransitionCondition(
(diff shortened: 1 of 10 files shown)


src/components/Sidebar/BranchTransition.tsx
```diff
@@ -219,6 +218,7 @@ const BranchTransition: FC<BranchTransitionProps> = ({ transition }) => {
 
             <TextField
               label="Criteria Display"
+              style={{ width: '100%' }}
```
radamson, line 221: set this in `styles.tsx` for consistency with where other styles are declared.

radamson (commented):
This PR is affected by the bug fixed by #57 so I'll retest once that fix is in and this is rebased.

One thing I did notice while testing is that newly created nodes are now automatically marked as `Medication`.  Was this intentional?  It wasn't immediately obvious that any of the code should have caused this, but I don't see this on master.

automed

KeeyanGhoreshi:
The reason it selects medication automatically is because the autocomplete component needs to be controlled, and having a value of "undefined" makes it uncontrolled when first rendered.  React complains when you go from uncontrolled -> controlled, so autocomplete needs a value that isn't undefined.  But autocomplete also complains when you make its value something that isn't included in the `options`, so we'd need to include an empty option to get default to.

But if we include an empty value, the dropdown gains an empty option, so it's got a blank space among the other, real options.

KeeyanGhoreshi:
ignore my previous comment, I just had a setting on the autocomplete component that was making everything way more difficult than it had to be.

dehall:
I looked briefly but it seems to crash anytime I do anything, so I think this needs a rebase to pull in Reece's fix.

Also the required dropdowns are no longer outlined in red when empty so there probably needs to be a CSS/className fix

Before:
image

After:
image

dehall:
One more minor display thing - the new component seems to have a lot more space around it than before. At a glance I couldn't tell what the difference was other than the new one seems to be wrapped in more layers of divs. Hopefully this is a one-line change but if it's not easy to figure out don't to spend a ton of time on it.

Before:
image


After:
image

noranda (approved):
This looks good! I particularly like how this allows us to clean up some code. :)

src/components/elements/DropDown/DropDown.tsx
```diff
@@ -12,47 +15,102 @@ interface DropDownProps {
...
+              {option.label}
+            </MenuItem>
+          ))}
+        </Select>
+      </>
+    );
+  };
+
   return (
     <FormControl variant="outlined" fullWidth>
```
noranda, line 112: This `FormControl` can be removed now, which is causing the extra margin issue since we already use `FormControl` in the `Autocomplete` component.

dehall (approved):

PR #60, opened 2020-10-21
Fix Undo stack when loading pathway by url

blangley28:
Previously when loading a pathway by url there was already something on the undo stack (https://mcode.github.io/pathway-builder/#/builder/eNPAgToBD/node/TumorSize). 

This was due to `criteriaSource` being set in the `PathwaysProvider`. The pathway would load but then `criteriaSource` would be set and update the pathway adding to the undo stack. 

Solution: Move the useEffect hook to the currentPathway. You can check this works by validating (https://mcode.github.io/pathway-builder/#/builder/eNPAgToBD/node/TumorSize):
1. The criteria source is still set in the sidebar 
2. There is nothing on the undo stack (can't undo)

Notes: While this fixes the issue (and removes the code from PatwhaysProvider which was going to happen in my integrate backend task anyway) there are numerous outstanding issues:
1. This is not very efficient as it will check the current pathway every time the pathway changes or the criteria changes. Note the pathway will not be reset so react will not re-render but it will still loop through all edges of the pathway on each change. Regardless this is still better than the previous solution which looped through every edge of every pathway on every change.
2. The reason this chunk of code exists is because criteriaIds change every time the page loads (and it loads the criteria from the static files). We should probably investigate a way of having deterministic ids every time (when pulling from the backend this should not be a problem because it is loading in a criteria object and not creating a criteria object from static files). The easiest way would be to have the app load json files of the criteria data model instead of re-importing every time.
3. Whenever criteria is deleted there is currently no check to make sure that it is safe. Here, like I am sure in other places in the builder, deleting the referenced item will result in a dangling reference.

### DIFF
--- a/src/components/CurrentPathwayProvider.tsx
+++ b/src/components/CurrentPathwayProvider.tsx
@@ -13,7 +13,7 @@ import useRefUndoState from 'hooks/useRefUndoState';
 import { usePathwaysContext } from './PathwaysProvider';
 import HotKeys from 'react-hot-keys';
 import { useCriteriaContext } from './CriteriaProvider';
-import { addCriteriaSourceToPathway } from 'utils/builder';
+import { updatePathwayCriteriaSources } from 'utils/builder';
 
 interface CurrentPathwayContextInterface {
   pathway: Pathway | null;
@@ -78,7 +78,7 @@ export const CurrentPathwayProvider: FC<CurrentPathwayProviderProps> = memo(({ c
   useEffect(() => {
     if (!pathway) return;
 
-    const { updated, newPathway } = addCriteriaSourceToPathway(pathway, criteria);
+    const { updated, newPathway } = updatePathwayCriteriaSources(pathway, criteria);
     if (updated) resetCurrentPathway(newPathway);
   }, [criteria, pathway, resetCurrentPathway]);
 
--- a/src/components/CurrentPathwayProvider.tsx
+++ b/src/components/CurrentPathwayProvider.tsx
@@ -12,6 +12,8 @@ import { Pathway } from 'pathways-model';
 import useRefUndoState from 'hooks/useRefUndoState';
 import { usePathwaysContext } from './PathwaysProvider';
 import HotKeys from 'react-hot-keys';
+import { useCriteriaContext } from './CriteriaProvider';
+import produce from 'immer';
 
 interface CurrentPathwayContextInterface {
   pathway: Pathway | null;
@@ -43,6 +45,7 @@ export const CurrentPathwayProvider: FC<CurrentPathwayProviderProps> = memo(({ c
     _resetPathway,
     _setPathway
   ] = useRefUndoState<Pathway | null>(null);
+  const { criteria } = useCriteriaContext();
   const { updatePathway } = usePathwaysContext();
 
   const undoPathway = useCallback(() => {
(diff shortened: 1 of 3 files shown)


src/components/CurrentPathwayProvider.tsx
```diff
@@ -71,6 +74,33 @@ export const CurrentPathwayProvider: FC<CurrentPathwayProviderProps> = memo(({ c
...
+    const newPathway = produce(pathway, draftPathway => {
+      Object.entries(draftPathway.nodes).forEach(([nodeIndex, node]) => {
+        node.transitions.forEach(({ condition }, transitionIndex) => {
+          // If a matching criteria does not already exist, try and find one
+          if (condition && !criteriaIds.includes(condition.criteriaSource as string)) {
+            const [library, statement] = condition.cql.split('.');
+            const criteriaSource = criteria.find(
+              crit => crit.elm?.library.identifier.id === library && crit.statement === statement
+            )?.id;
+            // Only update if a criteria source is actually found.
```
radamson, line 91: If no criteria source is actually found, will `immer.produce` still create a new object that is the same?

Previously the call to `setPathways` (`resetCurrentPathways` now in this PR) gets called even if there are no changes.  Does it make sense to have the the `resetCurrentPathway` and `produce` calls nested under the `if (criteriaSource)` so it only updates when a criteria source is available?

blangley28: Good catch - I think it does. The produce had to be brought out though because otherwise some weird recursion happened which crashed the app. I'll add a flag to only call `resetCurrentPathway` if a change happened.

radamson:
I'll give it a quick test again once it's rebased, but this looks good to me.

I am curious to see if this affects the export of multiple pathways.  Right now export relies on `criteriaSource` being properly set to gather and set the criteria CQL and this PR will only update those if the pathway being exported is set as the `currentPathway`.  Exporting multiple pathways does not rely on the `currentPathway` context so it seems likely that those pathways will not have updated `criteriaSource` fields.  It might not matter though if the CQL is already on the loaded pathway.

https://github.com/mcode/pathway-builder/blob/0a9ad3489dfc37bf45642b3a011d598bf4432434/src/utils/CaminoExporter.ts#L36-L38

src/utils/builder.ts
```diff
@@ -35,28 +35,73 @@ export function createNewPathway(name: string, description: string, pathwayId?:
...
+  return { updated, newPathway };
+}
+
+function addCriteriaSource(pathways: Pathway[], criteria: Criteria[]): Pathway[] {
+  const newPathways: Pathway[] = [];
+  pathways.forEach(pathway => {
+    const { newPathway } = addCriteriaSourceToPathway(pathway, criteria);
+    newPathways.push(newPathway);
+  });
+  return newPathways;
```
radamson, line 79: This looks like a good place to use `.map` instead of `.forEach...push`.

radamson, line 79: This looks like a good place to use `.map` instead of `.forEach...push`.

src/utils/builder.ts
```diff
@@ -35,28 +35,73 @@ export function createNewPathway(name: string, description: string, pathwayId?:
   };
 }
 
+interface AddCriteriaSourceInterface {
+  updated: boolean;
+  newPathway: Pathway;
+}
+
+export function addCriteriaSourceToPathway(
```
radamson, line 43: I feel like `updateCriteriaSources` would be a more meaningful name. ( or `updatePathwayCriteriaSources`?)

radamson (commented):
Library seems to be properly added now.

src/utils/builder.ts
blangley28: Good catch, much cleaner with a map

src/utils/builder.ts
blangley28: Changed to `updatePathwayCriteriaSources`

radamson (approved):
LGTM.

We should make sure to check this again while reviewing #65 since it removes `currentNode`.  I don't expect there to be any issues, but there might be an issue with directly using the broader context like there was here.

dehall:
One thing I'm noticing here is that the sample pathways are dependent on the sample criteria being present, as opposed to including the relevant libraries in the pathway itself (as would be done if it was exported). I'm wondering, if we updated the sample pathways so that they included the relevant libraries, would this whole chunk of code still be necessary? I don't think there's any other way for criteria to get "out of sync"

blangley28:
> I'm wondering, if we updated the sample pathways so that they included the relevant libraries, would this whole chunk of code still be necessary? I don't think there's any other way for criteria to get "out of sync"

I did check and this is certainly necessary (although I propose another solution below). Without running it on export the library field of the pathway is empty. To check if the useEffect hook is still necessary I exported a pathway (with the code block there to ensure the exported pathway is complete) and then re-imported. The imported pathway did not show criteria in the dropdown.

The reason this code exists is because on the static app criteria is loaded everytime the app is refreshed. This means the criteria id changes everytime. Two alternative solutions are to have the app load criteria (not cql files), like the database does in #63 or to have criteria id's deterministic

dehall:
Yeah, you're right. When the pathway is imported there always has to be some process to align the criteria in it. So I think we could potentially move this logic to the import process, because again there shouldn't be any other way for criteria to be out of sync (not counting deletes, that's a different kind of out of sync). 

But let's not spend too much more time on this. Frankly the undo stack already having something on it is not that big a deal. It's a little weird but not worth a major refactor to fix.

dehall (approved):

PR #61, opened 2020-10-22
Auth Page

dubem7:
To view the changes go to http://localhost:3000/#/

There is some code to be removed after review but the Modals would have otherwise been inaccessible.


src/components/Auth/Auth.module.scss
```diff
@@ -0,0 +1,150 @@
+@import '../../styles/variables';
+
+div {
```
radamson, line 3: This should be a `styles.tsx` if I'm not mistaken.  We've switched from using scss to makestyles/jss.

blangley28: Once you do that also update the colors use variables. There are a bunch already defined in the theme and if you want to add more they can be added there

src/components/Auth/AuthBody.tsx
```diff
@@ -0,0 +1,31 @@
...
+      <br />
+      <label className={styles.subText}>using mCODE. a standardized data </label>
+      <br />
+      <label className={styles.subText}> model around cancer</label>
+      <br />
+      <button className={styles.tryItButton}>Try it</button>
+      <br />
+      <span>
+        <img className={styles.images} src={github} alt="github" />
+        <img className={styles.images} src={email} alt="email" />
```
blangley28, line 25: There are fontawesome icons for these if we want to use them instead. I don't know if there's any difference

src/components/Auth/AuthBody.tsx
```diff
@@ -0,0 +1,31 @@
...
+const AuthBody: FC = () => {
+  const styles = useStyles();
+  return (
+    <>
+      <img src={example} alt="pathway" className={styles.exampleImage} />
+      <label className={styles.mainText}>A simple app to build</label>
+      <br />
+      <label className={styles.mainText}> customized clinical pathways</label>
+      <br />
+      <label className={styles.subText}>using mCODE. a standardized data </label>
```
blangley28, line 17: This should be a comma
```suggestion
      <label className={styles.subText}>using mCODE, a standardized data </label>
```

src/components/Auth/AuthFooter.tsx
```diff
@@ -0,0 +1,25 @@
...
+import mcode from 'mcode.png';
+import useStyles from './styles';
+
+const AuthFooter: FC = () => {
+  const styles = useStyles();
+  const { resetCurrentCriteria } = useCurrentCriteriaContext();
+
+  return (
+    <footer className={styles.footer}>
+      <Link to="/" className={styles.mitreButton} onClick={(): void => resetCurrentCriteria()}>
```
blangley28, line 15: Using a call back here (and in the `Link` below) instead of an arrow function will improve re-rendering the Link component

src/components/Auth/AuthHeader.tsx
```diff
@@ -0,0 +1,164 @@
...
+import NewPasswordModal from './NewPasswordModal';
+import PasswordResetModal from './PasswordResetModal';
+
+import logo from 'camino-builder-logo-light-bg.png';
+import useStyles from './styles';
+
+const AuthHeader: FC = () => {
+  const styles = useStyles();
+
+  const [openLogin, setOpenLogin] = useState<boolean>(false);
```
blangley28, line 18: Instead of defining like 6 different modals can it not just be one modal which renders differently based on the props?

src/index.js
```diff
@@ -17,4 +18,12 @@ if (process.env.NODE_ENV === 'development') {
   });
 }
 
-ReactDOM.render(<App />, document.getElementById('root'));
+ReactDOM.render(
+  <Router>
+    <Switch>
+      <Route exact path="/login" render={() => <Auth />}></Route>
+      <Route path="/" render={() => <App />}></Route>
```
blangley28, line 25: Personally I think the landing page should be `/`. That way when you just go directly to the base url you are prompted with the splash screen

src/components/Auth/styles.tsx
```diff
@@ -0,0 +1,146 @@
+import { makeStyles, Theme as AugmentedTheme } from '@material-ui/core/styles';
```
blangley28, line 1: Is this replacing the scss file? Once its been fully moved over be sure to delete that file

blangley28 (commented):
I know its a WIP so perhaps you already know this but I get the following error when I try and load http://localhost:3000/#/login

Screen Shot 2020-10-27 at 11 53 07 AM

src/components/Auth/Landing.tsx
```diff
@@ -0,0 +1,114 @@
...
+    setOpenReset(false);
+    setOpenSignup(false);
+    setOpenPasswordReset(false);
+    setOpenNewPassword(true);
+  }, []);
+
+  return (
+    <div className={styles.landing}>
+      <LandingHeader
+        openLogin={(): void => setOpenLogin(true)}
```
blangley28, line 64: There are a bunch of wdyr issues I think they're all related to using arrow functions in the props instead of using callbacks

src/components/Auth/LandingFooter.tsx
```diff
@@ -0,0 +1,37 @@
...
+import mcode from 'assets/mcode.png';
+import useStyles from './styles';
+
+const LandingFooter: FC = () => {
+  const styles = useStyles();
+  const { resetCurrentCriteria } = useCurrentCriteriaContext();
+
+  return (
+    <footer className={styles.footer}>
+      <a
```
blangley28, line 14: Clicking on either of these links gives 
`TypeError: resetCurrentCriteria is not a function`
and then does not load the website in the new tab. I think its because the routing happens in `index.js` so the App is never loaded (which is where the currentCriteriaContext is stored).

In fact, why does this component even need to resetCurrentCriteria()?

src/components/Auth/NewPasswordModal.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { FC, memo } from 'react';
+import { faLock } from '@fortawesome/free-solid-svg-icons';
+
+import Modal from '../elements/Modal';
+import TextInput from '../elements/TextInput';
```
blangley28, line 5: Use absolute imports here (and in all of the modals)
```suggestion
import Modal from 'components/elements/Modal';
import TextInput from 'components/elements/TextInput';
```

blangley28 (changes requested):
Since this also implements the /demo endpoint we will need to sync it with #64 which also implements the /demo endpoint but only for components in `components/StaticApp`

noranda:
Just a note that this should be good to review now. You should now go to `/` for the landing page and `/demo` for the current demo. You can also switch back and forth between them by clicking on the logos in the header.

src/components/Auth/LandingBody.tsx
```diff
@@ -0,0 +1,63 @@
...
+            className={styles.link}
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+            mCODE
+          </a>
+          , a standardized data model around cancer
+        </div>
+
+        <Button className={styles.tryItButton}>Try It</Button>
```
blangley28, line 34: Can we have this redirect to `/demo`. I know its technically part of the other task but just so we have a way to get to the actual builder.

blangley28: I know see your comment above, I guess there is a way so its up to you whether to implement this here or just let PATHWAYS-389 implement it

noranda: I think it's a good idea, so I implemented it. :)

src/components/Auth/NewPasswordModal.tsx
noranda: Oops, fixed all these imports.

src/components/Auth/LandingFooter.tsx
noranda: Oops again, fixed!

src/components/Auth/LandingFooter.tsx
```diff
@@ -0,0 +1,37 @@
...
+        href="https://mitre.org/"
```
blangley28, line 15: This still hangs for me. I think it needs to be 
```suggestion
        href="https://www.mitre.org/"
```

blangley28 (approved):

dehall:
This all looks good. The only thing I'm thinking is how will the routing structure work once we have user auth in place? I assume if someone is already logged in and goes to the root URL, they will either see their pathways/criteria there, or get redirected to the pathways/criteria list. I'm wondering if there's anything simple we could do here in preparation for that, and potentially avoid some future rework.

dehall:
2 more specific questions:
1 - Are the TODOs in this meant to be changed before merging this, or once the backend auth process is in place?
2 - Can we disable the login/signup buttons until the process is ready? Or possibly make it configurable to hide them? I don't expect we're going to have a public-facing backend server anytime soon, so the login/sign up buttons will never work on the github pages demo

dubem7:
> 1 - Are the TODOs in this meant to be changed before merging this, or once the backend auth process is in place? 


I was originally thinking before we merge. I just wanted those modals to be accessible while it was up for review.

> 2 - Can we disable the login/signup buttons until the process is ready? Or possibly make it configurable to hide them? I don't expect we're going to have a public-facing backend server anytime soon, so the login/sign up buttons will never work on the github pages demo


Do we just change the link in the readme to be /demo?

dehall:
Yeah just changing the link sounds fine. Then it's the same as how Camino works with the /demo endpoint.

I'll approve this but not merge it yet - go ahead and update the TODOs

dehall (approved):

PR #62, opened 2020-10-23
Support for importing CQL reference libraries

danlee1025:
You can now import a zip file, containing multiple CQL files.
    - Checks each cql file for criteria, if no criteria is present the file will be treated as a reference library.
    - Re-added `convertCQL` function to pass the libraries to the cql-to-elm webservice.
    - Added `cqlLibraries` property to criteria model to keep track of the reference libraries each criteria needs to run.

I still need to test output with cqf-ruler.

### DIFF
--- a/src/components/CriteriaProvider/CriteriaProvider.tsx
+++ b/src/components/CriteriaProvider/CriteriaProvider.tsx
@@ -15,7 +15,7 @@ import config from 'utils/ConfigManager';
 import useGetService from 'components/Services';
 import { ServiceLoaded } from 'pathways-objects';
 import { Criteria, BuilderModel } from 'criteria-model';
-import { convertBasicCQL, convertCQL, CqlLibrary, CqlObject } from 'engine/cql-to-elm';
+import { convertBasicCQL, convertCQL, CqlLibraries } from 'engine/cql-to-elm';
 
 interface CriteriaContextInterface {
   criteria: Criteria[];
@@ -47,13 +47,6 @@ const DEFAULT_ELM_STATEMENTS = [
   'Errors'
 ];
 
-const CQL_HELPER_LIBRARIES = [
-  'FHIRHelpers.cql',
-  'CDS_Connect_Commons_for_FHIRv400.cql',
-  'CDS_Connect_Commons_for_FHIRv401.cql',
-  'CDS_Connect_Conversions.cql'
-];
-
 function builderModelToCriteria(criteria: BuilderModel, label: string): Criteria {
   return {
     id: shortid.generate(),
@@ -66,7 +59,7 @@ function builderModelToCriteria(criteria: BuilderModel, label: string): Criteria
 function elmLibraryToCriteria(
   elm: ElmLibrary,
   cql: string | undefined = undefined,
-  cqlLibraries: CqlLibrary | undefined = undefined,
+  cqlLibraries: CqlLibraries | undefined = undefined,
   custom = false
 ): Criteria[] {
   // the cql-to-elm webservice always responds with ELM
@@ -116,18 +109,26 @@ function jsonToCriteria(rawElm: string): Criteria[] | undefined {
   return elmLibraryToCriteria(elm);
 }
 
-function cqlToCriteria(cql: string | CqlObject): Promise<Criteria[]> {
+function cqlToCriteria(cql: string | CqlLibraries): Promise<Criteria[]> {
   if (typeof cql === 'string') {
     return convertBasicCQL(cql).then(elm => elmLibraryToCriteria(elm, cql));
   } else {
     return convertCQL(cql).then(elm => {
       // Append library versions
-      Object.keys(elm.libraries).forEach(key => {
-        const cqlLibrary = cql.libraries[key];
-        const elmLibrary = elm.libraries[key] as ElmLibrary;
+      Object.keys(elm).forEach(key => {
+        const cqlLibrary = cql[key];
+        const elmLibrary = elm[key];
         if (cqlLibrary) cqlLibrary.version = elmLibrary.library.identifier.version;
       });
-      return elmLibraryToCriteria(elm.main as ElmLibrary, cql.main, cql.libraries);
+
+      // Loop through all elm libraries searching for criteria, use other libraries as dependencies
+      return Object.keys(elm)
+        .map(key => {
+          // Exclude current library from the list of cql libraries to pass to elmLibraryToCriteria
+          const { [key]: keyToExclude, ...dependencies } = cql;
+          return elmLibraryToCriteria(elm[key], cql[key].cql, dependencies);
+        })
+        .flat(1);
     });
   }
 }
@@ -146,7 +147,7 @@ export const CriteriaProvider: FC<CriteriaProviderProps> = memo(({ children }) =
     }
   }, [payload]);
 
-  const addCqlCriteria = useCallback((cql: string | CqlObject) => {
+  const addCqlCriteria = useCallback((cql: string | CqlLibraries) => {
     cqlToCriteria(cql).then(newCriteria => {
       if (newCriteria.length > 0) {
         setCriteria(currentCriteria => {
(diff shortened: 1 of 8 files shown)


src/engine/cql-to-elm.ts
```diff
@@ -1,9 +1,74 @@
...
 
 const url = config.get('cqlToElmWebserviceUrl');
 
+export interface CqlObject {
+  main: string;
+  libraries: Library;
+}
+
+export interface Library {
+  [name: string]: string; // should probably have an object for expected ELM structure.
```
blangley28, line 18: Is this interface for a CQL Library?

dehall: This came from Camino, I forget offhand if I added this as part of my current WIP PR or if it was already there. Just a wrapper for a mapping of `name of CQL library` -> `text of cql library`, it's not a cql or elm library by itself

danlee1025: Ahh yes it's used by `CqlObject`, I brought this over from camino. Should I re-name this to `CqlLibrary`?

blangley28: I think that would be helpful. Is the comment still relevant?

danlee1025: I removed the comment and the interface since it's not being used elsewhere and just explicitly declared in the `CqlObject` interface.

src/utils/cpg.ts
```diff
@@ -345,6 +345,19 @@ export class CPGExporter {
             };
 
             referencedDefines[transition.condition.cql] = libraryIdentifier.id;
+
+            if (criteriaSource?.cqlLibraries) {
```
dehall, line 349: I think we will need a similar change in `CaminoExporter.ts` as well.

dehall (approved):
Looks great. Two minor suggestions but I won't hold up the approval for those.

src/components/CriteriaProvider/CriteriaProvider.tsx
```diff
@@ -154,6 +167,21 @@ export const CriteriaProvider: FC<CriteriaProviderProps> = memo(({ children }) =
...
           } else if (file.name.endsWith('.cql')) {
             addCqlCriteria(rawContent);
+          } else if (file.name.endsWith('.zip')) {
+            const cqlObj: CqlObject = { main: '', libraries: {} };
+            const zip = await JSZip.loadAsync(file);
+            const zipFiles = Object.values(zip.files);
+            // Check for criteria in each of the cql files, add cql to libraries if no criteria
+            for (let i = 0; i < zipFiles.length; i++) {
+              const fileContents = await zipFiles[i].async('string');
+              const criteria = await cqlToCriteria(fileContents);
```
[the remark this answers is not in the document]
cmoesel: I'm not familiar enough w/ your use case to know why you need to determine the "main" library or what that means, but... I agree that you should send the whole set of files to the translation service.  The service needs access to all referenced libraries in order to translate correctly.

If you need to figure out what the "main" library is, I think your best bet is to analyze the dependencies of each library.  Any library that is _not a dependency of any other libraries_ is a candidate for being the "main" library.  If there is more than one, then I guess you'd need to establish a tie breaker -- maybe of those libraries, the one that itself depends on the most other libraries is the "main" library?  If you can save this analysis for _after_ translation it will be easier to pull out dependencies from the JSON ELM than it would be to pull them from the CQL text.

As for the Authoring Tool having predictably named helper libraries... you can't necessarily count on that.  Authors can upload arbitrary libraries as "External CQL" and those libraries will also be packaged into the zipped download file.

src/components/CriteriaProvider/CriteriaProvider.tsx
```diff
@@ -154,6 +173,24 @@ export const CriteriaProvider: FC<CriteriaProviderProps> = memo(({ children }) =
...
+            for (let i = 0; i < zipFiles.length; i++) {
+              const zipFile = zipFiles[i];
+
+              // Skip files that do not end with .cql
+              if (!zipFile.name.endsWith('.cql')) continue;
+              const fileContents = await zipFile.async('string');
+              if (CQL_HELPER_LIBRARIES.includes(zipFile.name)) {
+                cqlObj.libraries[zipFile.name] = { cql: fileContents };
+              } else {
+                cqlObj.main = fileContents;
```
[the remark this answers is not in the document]
danlee1025: Should be addressed now!

src/components/CriteriaProvider/CriteriaProvider.tsx
```diff
@@ -46,6 +47,12 @@ const DEFAULT_ELM_STATEMENTS = [
   'Errors'
 ];
 
+const CQL_HELPER_LIBRARIES = [
+  'FHIRHelpers.cql',
+  'CDS_Connect_Commons_for_FHIRv400.cql',
```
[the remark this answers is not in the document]
danlee1025: Added.

src/components/CriteriaProvider/CriteriaProvider.tsx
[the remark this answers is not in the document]
dehall: Thanks Chris! So after digging into this a little bit, this seems to have started from the pathway execution side of the house, and then that evolved into some incorrect assumptions about how cql-to-elm works. 
In the cql-execution `Library` class, it takes a single ELM object and optionally a Repository containing the dependencies. So there I think we do need to know a "main" library. (unless there's some way of passing them all in together, I admit I haven't looked too closely at that) But so what happened is we built a data model that expected a single main library and a list of dependency libraries.

On the pathway builder side here, we copy & pasted the same cql-to-elm conversion logic & data model because why reinvent the wheel. It may not be exactly the same, but here since we're showing the user individual criteria from the ZIP they uploaded, we still need to identify which library that specific criteria came from, and we can assume the rest are dependencies.

So I do think we can simplify things, though I don't know if it needs to be in this PR specifically:
 - update the cql-to-elm conversion process / data model to take just a list of CQL libraries instead of breaking out a "main" library
 - when we get a list of converted ELM back, loop through each one, and each criteria within it should be associated to the library it came from, with the rest as dependencies

@danlee1025 I'm good with this as-is, but if you think those 2 are easy enough to implement, it might be worth doing that here while we're thinking about it. Then we can get rid of the list of static helper files & related

src/components/CriteriaProvider/CriteriaProvider.tsx
```diff
@@ -154,6 +174,33 @@ export const CriteriaProvider: FC<CriteriaProviderProps> = memo(({ children }) =
...
+            // Check for criteria in each of the cql files, add cql to libraries if no criteria
+            for (let i = 0; i < zipFiles.length; i++) {
+              const zipFile = zipFiles[i];
+
+              // Skip files that do not end with .cql
+              if (!zipFile.name.endsWith('.cql')) continue;
+
+              const fileContents = await zipFile.async('string');
+              if (CQL_HELPER_LIBRARIES.includes(zipFile.name)) {
+                cqlObj.libraries[zipFile.name] = { cql: fileContents };
```
blangley28, line 190: I guess I'm a little confused - does the AT always produce only one cql? If someone uploads their own cql zip library this would fail. I guess I presumed the cql for the helper libraries would be known by the builder and then it would import the cql for any other files which are not in the list of helper files. Perhaps that is out of scope though

danlee1025: yea previous approach where I was separating out a "main" library and helper libraries was making a lot of assumptions. I updated the code to implement the approach Dylan mentioned in https://github.com/mcode/pathway-builder/pull/62#discussion_r518284866 so it should work even if we don't know the name of the helper libraries.

src/components/CriteriaProvider/CriteriaProvider.tsx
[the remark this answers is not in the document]
danlee1025: @dehall I think I updated to use this approach, lmk what you think.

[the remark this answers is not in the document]
dehall: Yeah looks great. If we really wanted to i think we could inspect the ELM to determine what the actual dependencies are, but that's overkill. 
This works with any zip containing CQL files, not just ones created in the CDS AT, so that's even better than I thought we'd get. 👍

blangley28 (approved):

PR #63, opened 2020-10-23
Backend for Saving Pathways

radamson:
Backend for saving, deleting and retrieving pathways, criteria and workspaces.


backend/README.md
```diff
@@ -0,0 +1,45 @@
+# Pathways Backend
```
blangley28, line 1: I know you were probably planning on expanding this but can we add instructions on how to install/run. Perhaps also link to the wiki page I made: https://github.com/mcode/pathway-builder/wiki/Backend-API

backend/src/index.ts
```diff
@@ -0,0 +1,24 @@
...
+mongoose.set('useFindAndModify', false);
+
+const app = express();
+const PORT = 8000;
+const payloadSizeLimit = '50mb';
+
+app.use(bodyParser.json({ limit: payloadSizeLimit }));
+app.use(cors());
+
+mongoose.connect('mongodb://localhost/test', { useNewUrlParser: true });
```
blangley28, line 20: I think the name of the database can be changed without causing any issues

backend/src/models/pathway.ts
```diff
@@ -0,0 +1,17 @@
+import mongoose from 'mongoose';
+
+const pathwaySchema = new mongoose.Schema({
```
blangley28, line 3: Have you looked into using the typescript interfaces with the mongoose schema. I think its more effort than its worth (especially for right now), but wanted to ask anyway.
https://medium.com/@tomanagle/strongly-typed-models-with-mongoose-and-typescript-7bc2f7197722

backend/src/routes/pathway.ts
```diff
@@ -0,0 +1,32 @@
...
+});
+
+router.put('/:id', (req, res) => {
+  if (req.params.id !== req.body.id) res.status(409).send('Pathway id does not match URL id');
+  Pathway.findOneAndUpdate(
+    { id: req.params.id },
+    req.body,
+    { overwrite: true, new: true },
+    (err, product) => {
+      if (err) res.send(err);
```
blangley28, line 17: What status code should the errors returns - 500?

blangley28 (commented):
Looks good - I don't see any major design issues which need to be addressed. Mostly just minor comments and questions for discussion.

backend/README.md
radamson: Ya I'll add those, just haven't gotten around to it yet.  Good call on the wiki.

backend/src/index.ts
radamson: I don't think it matters either so I'll change it to something better than `test`.  `pathways` perhaps?

backend/src/models/pathway.ts
radamson: I stumbled upon this while investigating the mongoose types issues, but I didn't take a serious look at it.  If we start doing some more serious business logic with the pathways or criteria in the backend then I think it would make sense.

backend/src/routes/pathway.ts
radamson: I think 500 makes sense for catching all the ones that aren't handled more specifically (like the 404's or 409's).

backend/src/index.ts
```diff
@@ -0,0 +1,31 @@
...
+import cors from 'cors';
+import config from './config';
+
+// Fix Deprecation Warnings
+mongoose.set('useCreateIndex', true);
+mongoose.set('useUnifiedTopology', true);
+mongoose.set('useFindAndModify', false);
+
+const app = express();
+const PORT = 8000;
```
blangley28, line 18: Since the config file exists perhaps this should go in there too

backend/src/routes/criteria.ts
```diff
@@ -0,0 +1,42 @@
+import { Router } from 'express';
+import Criteria from 'models/criteria';
+
+const router = Router();
+
+router.get('/', (req, res) => {
```
blangley28, line 6: From a style perspective how do you want to handle unused variables. I think the standard is to replace it with `_`

backend/src/routes/criteria.ts
```diff
@@ -0,0 +1,42 @@
...
+  Criteria.find({}, (err, product) => {
+    if (err) res.send('Error getting all Criteria');
```
blangley28, line 8: I noticed all of the errors do not send a status. I don't think the default will be an error code. It seems like there are a few ways to do it: https://expressjs.com/en/guide/error-handling.html
1. Using `res.send`, passing in the error status: `res.status(err.statusCode).send('msg')`
2. Passing the error to the default error handler which will send a response: `next(err)`

backend/src/routes/criteria.ts
```diff
@@ -0,0 +1,42 @@
...
+    else res.send(product);
```
blangley28, line 9: I think we should be explicit with all of our status codes... what is the default, 200?

backend/src/routes/pathway.ts
```diff
@@ -0,0 +1,42 @@
...
+
+router.put('/:id', (req, res) => {
+  if (req.params.id !== req.body.id) res.status(409).send('Pathway id does not match URL id');
+  Pathway.findOneAndUpdate(
+    { id: req.params.id },
+    req.body,
+    { overwrite: true, new: true, upsert: true },
+    (err, product) => {
+      if (err) res.send(err);
+      else res.status(201).send(product);
```
blangley28, line 21: Is there a way to check whether it was an update or a new document? For [PUT](https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html#sec9.6) it SHOULD return 200 on update (although since its a should its no biggie to return 201)

blangley28 (commented):
Even though we don't have users or workspaces yet for pathways or criteria I still think the routes should implement the query parameter endpoints (ie `/pathway?id=123`)

backend/src/routes/criteria.ts
radamson: I set all unhandled ones to 500 for now.  I'm not sure 1 is what we want because `err` in this case is a mongo error which don't match up to HTTP status codes.

backend/src/routes/pathway.ts
radamson: Return is now 201 on initial PUT, and 200 on all subsequent calls.

backend/src/handlers/crudHandler.ts
```diff
@@ -0,0 +1,67 @@
...
+
+interface Handler<T extends Document> {
+  model: Model<T>;
+  req: Request;
+  res: Response;
+}
+
+const getAllHandler = <T extends Document>({ model, res }: Handler<T>): void => {
+  model.find({}, (err, product) => {
+    if (err) res.status(500).send('Error getting all pathways');
```
blangley28, line 12: Is there a way to make the message here use the type instead of always saying 'pathways'?

blangley28 (approved):
LGTM

dehall:
I haven't looked closely at the code yet, will do that next, but is having the `backend` subfolder going to complicate our travis config? We should have CI run `yarn lint` and `yarn test` on the backend project too.

backend/src/models/criteria.ts
```diff
@@ -0,0 +1,18 @@
+import { Schema, model } from 'mongoose';
+
+const criteriaSchema = new Schema({
+  id: String,
```
dehall, line 4: The API document had something slightly different for the schemas - here the schema defines the fields of the object as we know it, but the doc had these wrapped up in another class that just had id, user, and "the actual object" fields.
https://github.com/mcode/pathway-builder/wiki/Backend-API#collection-schemas

I'm assuming we always want to receive/return the regular object on webservice calls, but the wrapped object might make things easier for user/workspace grouping. Thoughts? (cc: @blangley28 )

radamson: I updated the schemas/models so that all the models are now of the form

```
metadata: Object,
value: Object
```

where `value` is the actual pathway model (e.g. Criteria, Pathway, Workspace) and `metadata` is a catch all for things like `userid` for the user that owns it.  I didn't flesh the metadata out in any detail because right now we don't have any routes that provide that information.  I could, but we'd probably want to think about updating the API so we can actually drive that.

I'll update the wiki docs now.

blangley28 (approved):
Still approve these changes. Everything looks good, api still works, and tested against my backend task

backend/src/models/criteria.ts
dehall: Would it be possible to promote the ID field to the top-level of each model as well? Maybe it's not as big a deal as I think but it seems like the ID is fundamental enough that it should exist at the top level

radamson: I can do that.  Originally I had it added as a virtual field, but since those aren't saved to the database we can't query based on them.  I should be able to duplicate the field in the database so that we can search off it though.

radamson: Added the id to the top level and updated the index & queries to be based on it.

dehall (approved):

PR #64, opened 2020-10-23
Integrate the Backend 

blangley28:
Integrating the backend from [PR-63](https://github.com/mcode/pathway-builder/pull/63) into the react app. Utilizes the [React Query](https://react-query.tanstack.com/) library which uses a cache to synchronize the app state with the server state. 

To test you will need to setup the backend. Details are included in the [backend README](https://github.com/mcode/pathway-builder/tree/integrate-backend/backend). I added a reset endpoint to the pathway and criteria routes which will purge the collections and repopulate with sample data. You can find all of the scripts in the backend folder of [this postman collection](https://www.getpostman.com/collections/95c85ccd540abefcaddd).

Also includes a `/demo` endpoint which uses the static files


backend/package.json
```diff
@@ -7,6 +7,7 @@
   "license": "Apache 2.0",
   "private": true,
   "dependencies": {
+    "@types/cors": "^2.8.8",
```
radamson, line 10: This is already added as a `devDependency`

src/components/App.tsx
```diff
@@ -16,45 +14,50 @@ import { CurrentNodeProvider } from './CurrentNodeProvider';
 import { CurrentCriteriaProvider } from './CurrentCriteriaProvider';
 import { SnackbarProvider } from './SnackbarProvider';
 import { CriteriaBuilderProvider } from './CriteriaBuilderProvider';
+import { QueryCache, ReactQueryCacheProvider } from 'react-query';
+import StaticApp from './StaticApp/StaticApp';
```
radamson, line 18: Add an `index.ts` in `./StaticApp` so this can be imported as `./StaticApp` similar to how we do it for other components contained in a directory

src/components/CurrentPathwayProvider.tsx
```diff
@@ -68,8 +72,41 @@ export const CurrentPathwayProvider: FC<CurrentPathwayProviderProps> = memo(({ c
...
+    if (pathway) {
+      updatePathway(pathway).then(() => cache.invalidateQueries('pathways'));
+    }
+  }, [cache, pathway]);
+
+  // Update criteriaSource if criteria change
+  useEffect(() => {
+    if (!pathway) return;
+
+    let updated = false;
```
radamson, line 84: More a reminder to myself than anything else, but I assume this will use `updatePathwayCriteriaSources` once #60 is merged.

src/components/StaticApp/ActionButton/ActionButton.tsx
```diff
@@ -0,0 +1,71 @@
+import React, { FC, memo } from 'react';
```
radamson, line 1: Do we actually need to duplicate the `ActionButton` in `static` app or can we just use `component/ActionButton`.  It doesn't look different at a glance.

blangley28: Short answer: no.
Long answer: I duplicated all of the components just because a) I didn't want to take the time to figure out which files could stay and which files had to change and b) I don't know how long this temporary fix will be live for, and I didn't want to cause possibly unforseen issues on the github demo because the static app was referencing a file which changed to be non-static. 

I can certainly go through and update these if its worth our time but for all future PRs until the useLocalStorage task is merged, we will need to be extra vigilant about whether or not the changes break the demo

src/components/CurrentPathwayProvider.tsx
blangley28: Yes, I had to make at least some change since the PathwayProvider has been deleted. Instead of trying to copy over all of the changes as they occur I am planning on waiting until #60 is merged.

blangley28: After reviewing Dylan's comment on #60 this shouldn't be necessary in this component (although the change needs to be carried over into `StaticApp/CurrentPathwayProvider`

blangley28:
Is anyone else having issues adding a new node? It seems like the order of the useEffect hooks isn't always the same so sometimes there is unexpected behavior (where the node is added but then immediately overwritten - you can check because no node appears but the undo button allows you to go back and see the new node there)

blangley28:
So I figured out the issue with the new node being added but now that new node isn't being highlighted as current. I know why but I need to figure out how to solve the problem with reactQuery

src/components/StaticApp/ActionButton/ActionButton.tsx
blangley28: Updated to only include required components

dehall:
Not sure if I am doing something wrong but the import pathway feature doesn't seem to import the associated criteria anymore. I haven't looked closely at the code yet to try to figure out why

Sample pathway to test with (works fine on master/github demo)  [New Pathway(1).json.txt](https://github.com/mcode/pathway-builder/files/5495199/New.Pathway.1.json.txt)

blangley28:
> the import pathway feature doesn't seem to import the associated criteria anymore. I haven't looked closely at the code yet to try to figure out why

Good catch - its related to that useEffect hook in `currentPathwayProvider` from my last PR. I removed it but I guess it still needs to run on import

blangley28:
@dehall fixed the issue where criteria was not being loaded. Note if there is criteria it needs to import it takes a second or two for the pathway to appear in the list. This is because it needs to add all the criteria first. If I add the pathway before the criteria is converted (and given an id) then the user could click on the pathway before the database is fully updated. Since the app uses a context to provide the current pathway and not the react query, there is no way to invalidate the request and force a refresh. There's a few possibilities here (which hopefully can be separate tasks):
1. Immediately return a criteria source and then just add the criteria once it is created
      -> Not as straightforward as it sounds (but not as complex as option 2). Would require reworking how the loading pathway criteria is handled.
2. Remove the currentPathwayProvider in favor of a usePathway() hook 
      -> Pretty large overhaul as this is responsible for synchronizing to the database. I'm also not sure how undo/redo would work without this context

dehall:
No worries about a small delay. If we could easily add a little spinny loading image that would be great, or somehow indicate that the pathway is loaded but not ready yet, but even those would be more work than it's worth here.

I'm testing through it now and looks good so far, will let you know if I see anything else unexpected.

blangley28:
>If we could easily add a little spinny loading image that would be great, or somehow indicate that the pathway is loaded but not ready yet

I can do this but want to loop in @noranda for advice. I don't think it would make sense to hide the entire pathways table with a loading screen. Perhaps replace the the import modal with a loading screen and close it when its imported? I don't think that would be too difficult to do

dehall:
One more thought, we don't care too much about fault-tolerance at this point, because that will complicate things a lot. 
But I just tested out two backend stability situations: 
1) load the UI without the backend running -- here the lists just show a spinning loading image. If there's a way it could indicate that the server isn't responding that would be good. Not a priority though, let's make that a backlog item. 

2) load up a pathway while the backend is running, then stop the backend. This time I got an error popup "ERROR: Could not load pathway. Redirecting to Pathway List screen" which isn't ideal because now the last changes I just made didn't save, and I don't have the ability to export it and save off my changes separately. Rather than get too fancy, I would say just remove the redirect and show an error that tells the user "everything is borked, save off this pathway just in case and restart"? Though based on where that message exists in the code, I don't think this is where you expected it to pop up. Also not a super high priority, but maybe there's an easy solution?

blangley28:
Added alerts for when an error with the backend occurs. If the backend goes down while editing a pathway you can still download it (no redirect). 

Updated the pathways import modal to show a little loading icon while the import is loading. I think its pretty neat

dehall (approved):
Looks great. I'll note the backend is included here, and I didn't really look at that at all, since that's its own separate PR and this is only going into the backend branch, not master

KeeyanGhoreshi (approved):
lgtm

dehall:
One thing I just noticed when testing out the default criteria label PR, based off this one, is that if you switch between pathways, when you load up the second one it shows a flash of the first one before the second one loads. Not a showstopper but might be worth looking into.

blangley28:
So I due to all the complications what I just did was as follows:
1. Merge master into this branch
2. Manually merge some changes from files which were deleted
2. Squashing commits did not work so I reset the branch to the merged version
3. Rebased off master

I think everything is sorted now but can I get another review please to make sure everything from previous PRs is included. @danlee1025 would you check your CQL stuff all works too? I deleted one of the providers you were working in so I had to manually copy over the changes.

dehall:
I just briefly checked and things looked good. 
For some reason GitHub is expecting a 3rd travis check though, so I'm not sure what that's about

danlee1025 (approved):

PR #65, opened 2020-10-28
remove current node provider

KeeyanGhoreshi:
The cursor bug is a symptom of a larger issue where the `currentNode` is not in sync with the same node in the pathway.  The two versions of the node being out of sync briefly causes the cursor to jump when editing text fields that draw their value from the `currentNode`.

This PR removes the `currentNodeProvider` altogether and instead the nodes are passed the current node from the top level Builder component.  The nodes also access the current node by parsing the URL the same way the BuilderRoute did originally.   Since `currentNodeProvider` provided both the `currentNode` and a `currentNodeRef`,  the nodes compensate by replacing `currentNodeRef` with the version in the current pathway. 

This fixes the cursor bug as a consequence.


blangley28 (commented):
Looks good, bug fixed for me. The following block appears a lot.
```
const { nodeId } = useParams();
const currentNodeId = decodeURIComponent(nodeId);
const currentNodeStatic = pathway?.nodes[currentNodeId];
```
Is there anyway this can be defined once and then called by each component?

KeeyanGhoreshi:
I wanted to somehow bring that block out and make it available to all the components but I'm not sure how to do so.

blangley28:
> I wanted to somehow bring that block out and make it available to all the components but I'm not sure how to do so. My original idea was to make the currentNodeProvider simply provide that value but that did not work.

Would it work to make a new hook (`useCurrentNodeStatic`) which takes in the `pathway` as an argument?

dubem7 (approved):
This fixes the issue, LGTM.

KeeyanGhoreshi:
> Would it work to make a new hook (useCurrentNodeStatic) which takes in the pathway as an argument?

The problem is the `useParam` for the nodeId, since it needs to be inside the Router element.

src/components/Builder/Builder.tsx
```diff
@@ -40,7 +49,7 @@ const Builder: FC = () => {
       {pathway && (
         <div className={styles.main}>
           <MuiThemeProvider theme={theme}>
-            <Sidebar />
+            <Sidebar currentNode={currentNode} />
```
blangley28, line 52: Is there a reason `Sidebar` and `DagreGraph` can't compute the `currentNode` themselves? They can get the pathway via the context and the nodeId via params

blangley28:
Oh oops - I thought I pushed that to a different branch... sorry. I created those hooks and removed the currentnode props from the sidebar in this branch: https://github.com/mcode/pathway-builder/tree/cursor-bug-ben

src/components/Builder/Builder.tsx
KeeyanGhoreshi: the passed down version of currentNode is not the same as the version of currentNode retrieved from the pathway inside that component.

KeeyanGhoreshi:
on your branch the bug actually reappears in the sidebar's node editor inputs, because the static current node is not the same as the current node being passed through in the props.

src/components/Sidebar/ActionNodeEditor.tsx
```diff
@@ -56,31 +57,31 @@ const ActionNodeEditor: FC<ActionNodeEditorProps> = ({ changeNodeType }) => {
 
   const changeCode = useCallback(
     (event: ChangeEvent<{ value: string }>): void => {
-      const currentNode = currentNodeRef.current as ActionNode;
+      const currentNode = currentNodeStatic as ActionNode;
```
blangley28, line 60: Would it be easier to just define currentNodeStatic as ActionNode instead of doing this in every callback?

KeeyanGhoreshi: ya, good point

src/hooks/useNodeId.ts
```diff
@@ -0,0 +1,8 @@
+import { useParams } from 'react-router-dom';
+
+const useNodeId = () => {
```
dehall, line 3: Travis failed on the last commit, I think you just need to add the return type here

src/hooks/usePathwayId.ts
```diff
@@ -0,0 +1,8 @@
+import { useParams } from 'react-router-dom';
+
+const usePathwayId = () => {
```
dehall, line 3: ...and here

blangley28 (approved):

PR #66, opened 2020-11-05
Default criteria display

blangley28:
All criteria now has a default display. If the transition condition does not have a description it will use the default display instead

### DIFF
--- a/src/components/Sidebar/BranchTransition.tsx
+++ b/src/components/Sidebar/BranchTransition.tsx
@@ -92,11 +92,12 @@ const BranchTransition: FC<BranchTransitionProps> = ({ transition, currentNode }
 
       const selectedCriteria = criteria.find(c => c.id === criteriaId);
       if (!selectedCriteria) return;
+      console.log(selectedCriteria);
       const newPathway = setTransitionCondition(
         pathwayRef.current,
         currentNodeStatic.key,
         transitionRef.current.id,
-        transitionRef.current.condition?.description || '',
+        transitionRef.current.condition?.description || selectedCriteria.display,
         selectedCriteria
       );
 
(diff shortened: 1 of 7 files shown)


src/utils/criteria.ts
```diff
@@ -38,6 +38,7 @@ export function elmLibraryToCriteria(
     return {
       id: shortid.generate(),
       label: `${labelTitle}: ${statement.name}`,
+      display: statement.name,
```
dehall, line 41: I was kind of hoping we could do this without adding another field with duplicate content. It looks like both here and in `builderModelToCriteria` below the `statement` field gets the same value as this new `display` field. Is there any reason we can't use that field here?

blangley28: Oh yea, I didn't realize it was the same... I initially made it a new label though because I envisioned having some way to edit it from the PathwayList but if we don't want to give the user that possibility I can just pull the display from the statement

dehall: I mean, it's a nice feature idea but it would be pretty low in the priority list. And it is nice to have a separate field name that clearly indicates it's for display 

But if we stuck with `statement` would that reduce this to just the 2 lines of:
```
transitionRef.current.condition?.description || selectedCriteria.display
```

becoming
```
transitionRef.current.condition?.description || selectedCriteria.statement
```

blangley28: Yes

src/components/Sidebar/BranchTransition.tsx
```diff
@@ -92,11 +92,12 @@ const BranchTransition: FC<BranchTransitionProps> = ({ transition }) => {
 
       const selectedCriteria = criteria.find(c => c.id === criteriaId);
       if (!selectedCriteria) return;
+      console.log(selectedCriteria);
```
danlee1025, line 95: Looks like a leftover `console.log` we want to remove.

danlee1025:
This looks pretty good to me, did you plan to implement the change that Dylan proposed?

dehall:
> This looks pretty good to me, did you plan to implement the change that Dylan proposed?
We talked about that on Slack and I said it's not worth it. So I'll mark that as resolved.

danlee1025 (approved):

dehall (approved):

PR #67, opened 2020-11-09
Implement Try It Button

blangley28:
Try it button now redirects to the /demo endpoint which shows a banner explaining nothing will be saved. The banner links to the login and signup modals which have been refactored to allows them to be easily used in the two different components.

### DIFF
--- a/src/components/StaticApp/Header/Header.tsx
+++ b/src/components/StaticApp/Header/Header.tsx
@@ -14,19 +14,19 @@ import useStyles from './styles';
 import { useAuthModalContext } from 'components/Auth/AuthModalProvider';
 
 const DemoBanner: FC = () => {
-  const styles = useStyles();
+  const styleOverrides = useStyles();
   const { openLoginModal, openSignupModal } = useAuthModalContext();
 
   return (
-    <div className={styles.root}>
-      <FontAwesomeIcon icon={faExclamationCircle} className={styles.icon} />
+    <div className={styleOverrides.root}>
+      <FontAwesomeIcon icon={faExclamationCircle} className={styleOverrides.icon} />
       You are currently viewing syntheic demo data. Any changes made will not be saved. To create
       your own pathways,{' '}
-      <span className={styles.underline} onClick={openLoginModal}>
+      <span className={styleOverrides.underline} onClick={openLoginModal}>
         login
       </span>{' '}
       or{' '}
-      <span className={styles.underline} onClick={openSignupModal}>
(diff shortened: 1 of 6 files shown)


danlee1025:
This looks pretty good, just one question. Is clicking on the the camino logo supposed to bring us back to the landing page?

src/components/StaticApp/Header/Header.tsx
```diff
@@ -10,6 +10,29 @@ import { useThemeToggle } from 'components/ThemeProvider';
...
+
+  return (
+    <div className={styles.root}>
+      <FontAwesomeIcon icon={faExclamationCircle} className={styles.icon} />
+      You are currently viewing syntheic demo data. Any changes made will not be saved. To create
+      your own pathways,{' '}
+      <span className={styles.underline} onClick={openLoginModal}>
+        login
+      </span>{' '}
+      or{' '}
```
radamson, line 28: Does the space have to be in `{' '}`?  Maybe its doing something I don't understand.

src/index.js
```diff
@@ -11,6 +11,7 @@ import ThemeProvider from 'components/ThemeProvider';
 
 import './styles/index.scss';
 import StaticApp from 'components/StaticApp';
+import { AuthModalProvider } from 'components/Auth/AuthModalProvider';
```
radamson, line 14: any reason `AuthModalProvider` export isn't added to `Auth/index.ts`?  I don't have a strong opinion on it, but some of the components in the directory are included and some aren't

src/components/StaticApp/Header/Header.tsx
blangley28: Everytime I deleted it the formatter just added it back. I removed it in a different editor and replaced them with spaces but the spaces don't render - so I guess it is required.
Screen Shot 2020-11-12 at 2 12 38 PM

src/index.js
blangley28: Nope - added to `Auth/index.ts`

radamson (approved):
Tested the `/demo` and `/builder` pages and they both were working as expected.  If you switch to another page via the browser URL (e.g. `/demo` to `/builder`) with either the sign-up or login modals open they stay open, but you can still close them so its not a big deal

blangley28:
> Is clicking on the the camino logo supposed to bring us back to the landing page?

Talked with Reece and we decided clicking the logo on the static app should bring the user back to the landing page but not from the backend builder. Updated with this functionality.

danlee1025 (approved):

PR #68, opened 2020-11-17
updated react

KeeyanGhoreshi:
Updated to react 17, doesn't seem to have caused any errors.  I tested all the functionality I could and couldn't find anything.  React 17 does seem to contain very few breaking changes that would cause problems, according to their website, so it checks out.


package.json
```diff
@@ -33,8 +33,8 @@
     "jszip": "^3.5.0",
     "moment": "^2.25.3",
     "node-sass": "^4.12.0",
-    "react": "^16.10.1",
-    "react-dom": "^16.10.1",
+    "react": "17.0.0",
```
radamson, line 36: [React on npm](https://www.npmjs.com/package/react) went to 17.0.1 a month ago.  Not sure whats in the patch release, but we might as well upgrade to that.  Also if there is no compelling reason to have the exact version here, I'd maintain the [caret](https://docs.npmjs.com/cli/v6/using-npm/semver#caret-ranges-123-025-004) that was previously present.

radamson (approved):

dehall (approved):

PR #70, opened 2020-11-23
Convert CQL to ELM only once

blangley28:
Update the builder so the action node cql is only converted to ELM at export time. It then proceeds to add this to the elm. Since the conversion can take a couple seconds I also added a little modal showing the application is working on exporting the pathway.


blangley28:
Closed in favor of #73

PR #71, opened 2020-11-25
Fixes for some cql issues

dehall:
Fixes for some issues identified when exporting pathways into Camino:

1. Identifiers that start with numbers aren't valid in CQL, so keys for nodes and transitions are now prepended with 'N' or 'T'.
2. Identifiers may contain spaces, so in the construction of the "main" CQL, wrap the statement name in quotes. This may make 1 ^ OBE, now that I think about it, but I didn't try it out. At the same time I updated the default criteria CQL files to have names with underscores instead of dashes, just to help minimize bugs
3. Libraries were being referenced in that "main" CQL as their entire text, causing errors.
4. Update everything to FHIR 4.0.1 instead of 4.0.0 so that it's consistent and will work with cqf-ruler. (note there are some test fixtures that still reference 4.0.0 - I can update these if desired but the fact that they use 4.0.0 doesn't really matter internally)
5. Importing FHIRHelpers in a zip brought in the "ToBoolean" function which can't actually be a meaningful criteria. Updated the importer to exclude functions
6. Updated the demo criteria to import pre-converted criteria objects instead of importing raw CQL. I like this less than using the real import feature, but it was a lot less work than updating the Service interface to support reading zips. These pre-converted criteria include FHIRHelpers. Also updated the backend default criteria, note the backend default criteria and frontend demo criteria files are basically identical. 
   - Note: I did not update the default pathways here since there are still some kinks to work out in general. My suggestion is let's wait until we have a good baseline then re-create them fresh, rather than continuing to update the old hand-created pathway files.

Base branch for this is `cql-convert`


src/utils/CaminoExporter.ts
```diff
@@ -81,7 +81,6 @@ export class CaminoExporter {
                     cql: libCql.cql,
                     version: libCql?.version || ''
                   };
-                  referencedDefines[libCql.cql] = libName;
```
blangley28, line 84: Not sure I understand why this is being removed

blangley28 (commented):
Nice changes. Should all the criteria be updated to 4.0.1 as well? See `public/static/cql` and `backend/src/defaultData/criteria.ts`

src/utils/CaminoExporter.ts
dehall: This is the fix for # 3 in my list above. `libCql.cql` is the entire content of the referenced cql library. And then look at how `referencedDefines` is used in the `constructCqlLibrary` function:

```
  const definesList = Object.entries(referencedDefines).map(
    ([name, srcLibrary]) => `define "${name}": ${srcLibrary}."${name}"\n\n`
  );
```
This is meant to just re-map definitions from a dependency library, so like `define "T > 2": TNM."T>2"`

And I'm pretty sure everything we need is included elsewhere, so this line can just be removed.
The relevant bits to make that remapping work for this specific criteria are set a few lines above:
```
referencedDefines[transition.condition.cql] = libraryIdentifier.id;
```
and the relevant library gets included by this part:
```
includedCqlLibraries[libName] = {
  cql: libCql.cql,
  version: libCql?.version || ''
};
```

dehall:
Yes the criteria will be updated to 4.0.1 alongside whatever process I use to include FHIRHelpers

dehall:
Update: I'll definitely update the backend default criteria here, but I'm thinking let's hold off on updating the default pathways until after #70 is done. The current pathways were manually created a long time ago and have been manually updated since then. Let's re-create them from scratch once we think we have a good baseline.

dehall:
Closing this due to the target branch changing & rebasing. Watch for #74

PR #74, opened 2020-12-02
Fixes for some CQL issues

dehall:
Fixes for some issues identified when exporting pathways into Camino:
(NOTE: to test this in Camino, use the `builder-model` branch there)

1. Identifiers that start with numbers aren't valid in CQL, so keys for nodes and transitions are now prepended with 'N' or 'T'.
2. Identifiers may contain spaces, so in the construction of the "main" CQL, wrap the statement name in quotes. This may make 1 ^ OBE, now that I think about it, but I didn't try it out. At the same time I updated the default criteria CQL files to have names with underscores instead of dashes, just to help minimize bugs
3. Libraries were being referenced in that "main" CQL as their entire text, causing errors.
4. Update everything to FHIR 4.0.1 instead of 4.0.0 so that it's consistent and will work with cqf-ruler. (note there are some test fixtures that still reference 4.0.0 - I can update these if desired but the fact that they use 4.0.0 doesn't really matter internally)
5. Importing FHIRHelpers in a zip brought in the "ToBoolean" function which can't actually be a meaningful criteria. Updated the importer to exclude functions
6. Updated the demo criteria to import pre-converted criteria objects instead of importing raw CQL. I like this less than using the real import feature, but it was a lot less work than updating the Service interface to support reading zips. These pre-converted criteria include FHIRHelpers. Also updated the backend default criteria, note the backend default criteria and frontend demo criteria files are basically identical. 
   - Note: I did not update the default pathways here since there are still some kinks to work out in general. My suggestion is let's wait until we have a good baseline then re-create them fresh, rather than continuing to update the old hand-created pathway files.

Base branch for this is `cql-convert-build-criteria`. Replaces #71


blangley28 (commented):
CQL/ELm stuff looks good. I'm not sure how this happened but it seems the default criteria label is broken on this branch but works on the `cql-convert-build-criteria` branch.

dehall:
Yeah I realized that too, I need to update that. I think it's just that the default criteria don't have labels, not a code issue

blangley28 (approved):

danlee1025 (approved):
LGTM, tested the exported pathways on the `builder-model` branch of Camino and it loaded fine.

PR #75, opened 2020-12-04
Use cql-translation-service-client library

radamson:
Updates the pathway builder to use the [`cql-translation-service-client library`](https://github.com/mcode/cql-translation-service-client).

`convertCQL` and `convertBasicCQL` implementations are now delegated to the library.  Some type assertions are used because the builder library and ctsc libraries provide their own typings for things like `ElmLibrary`.

I also removed the test for `convertBasicCql` because it relied on mocking the `global.fetch`, and the ctsc library doesn't use fetch.  I wasn't sure the best way to change the mock to make this work without looking into the libraries implementation, so I figured for now I'd just delete it.

[Some](https://stackoverflow.com/questions/63331421/export-a-type-after-its-declaration-on-typescript-3-8) [resources](https://github.com/babel/babel/issues/6065#issuecomment-321121848) on the gymnastics used to re-export `CqlLibraries`.

**WIP until https://github.com/mcode/cql-translation-service-client/pull/4 is merged and a new version is published to NPM**


src/engine/cql-to-elm.ts
```diff
@@ -1,72 +1,24 @@
...
-      'Content-Type': 'application/cql',
-      Accept: 'application/elm+json'
-    },
-    body: cql
-  }).then(elm => elm.json());
+export function convertBasicCQL(cql: string): Promise<BuilderElmLibrary> {
+  return client.convertBasicCQL(cql) as Promise<BuilderElmLibrary>;
 }
+
+export type CqlLibraries = _cl;
```
blangley28, line 24: Is there a reason we do this instead of just having other files `import { CqlLibraries } from 'cql-translation-service-client';`

radamson: The main benefit to re-exporting here is that we don't need to change the other files and they can remain only dependent on this one file, and not this file and the exported type. Everything in this file is essentially just a wrapper around the cql-translation-service client methods, with updated type assertions, and types. This files real role now is the hide that dependency and do the necessary type assertions in one place.

I should note that the types provided by `cql-translation-service-client` are equivalent to the types in the pathway-builder, but still actually different so it needs the type assertion. There might be a better way to share and use the types between these two, but I didn't dig in too much into that and this seemed the least intrusive.

blangley28: Makes a lot of sense! LGTM - once the the client is updated you just need to change the dependencies right?

radamson: Yup, waiting for https://github.com/mcode/cql-translation-service-client/pull/4 to get in and an updated package published to NPM.

I converted it to Draft so it's harder to accidentally merge 🙂 .

dehall:
FYI the client has been updated to 0.6.0 on NPM https://www.npmjs.com/package/cql-translation-service-client

blangley28 (approved):
LGTM

dehall (approved):

